PR #76, opened 2022-05-18
fix: update config.ts to reflect config/default.json

andreasmwenzel:


andreasmwenzel:
this may bork the build if it has the wrong config file set up

PR #79, opened 2022-05-22
Feat/image gen microservice

andreasmwenzel:


core/src/image-generation/image-generation.controller.ts
```diff
@@ -0,0 +1,14 @@
+import {Controller, Get, Param} from "@nestjs/common";
+
+import {ImageGenerationService} from "./image-generation.service";
+
+@Controller("image-gen")
```
0c370t, line 5: GraphQL, not Rest please

0c370t (changes requested):
Overall, very well done. One tweak and it is good to go

core/src/image-generation/image-generation.controller.ts
andreasmwenzel: Does this actually need a controller? I meant to ask that. I threw this up for really easy testing, but we'll this be called from the web or just imported by the scrim module and called from within nest?

0c370t: That's a good question, and something that we probably need to explore more in the future. Leave it in for now.

0c370t (approved):

PR #102, opened 2022-06-05
feat: retry on 500 failures

zachstence:
- Make response checkers more safe by testing `status_code in response`, and using `dict.get` instead of `dict["str"]`
- Add retry loop to parse section
- Retry on all unknown errors (including 500s from ballchasing)


0c370t:
Task linked: [CU-2cynjzb Replay Parsing retry on 500](https://app.clickup.com/t/2cynjzb)

0c370t (approved):

PR #108, opened 2022-06-07
Hide manage button for non-admins

0c370t:


0c370t:
Task linked: [CU-2echnwe Hide manage button from nav when a user is not an admin](https://app.clickup.com/t/2echnwe)

bwpikaard (approved):

PR #109, opened 2022-06-07
Queue Ban Notifications

bwpikaard:


0c370t:
Task linked: [CU-2jttugk Notify Users Who Get Queue Banned](https://app.clickup.com/t/2jttugk)

common/src/service-connectors/core/schemas/MemberRestriction.schema.ts
```diff
@@ -25,4 +30,4 @@ export const MemberRestrictionSchema = z.object({
     memberId: z.number(),
 });
 
-export type MemberRestrictionSchemaType = z.infer<typeof MemberRestrictionSchema>;
```
gankoji: This name conflicts at point of use elsewhere in core (there is already a type `MemberRestriction` from the database model). hence why I named it wonky. Could be a better name, but it really shouldn't be `MemberRestriction`

gankoji (changes requested):

gankoji (approved):

microservices/notification-service/src/member/member.event-subscriber.ts
```diff
@@ -0,0 +1,31 @@
...
+        await this.eventsService.subscribe(EventTopic.MemberRestrictionCreated, false).then(obs => {
+            // eslint-disable-next-line @typescript-eslint/no-misused-promises
+            obs.subscribe(this.onMemberRestrictionCreated);
+        });
+    }
+
+    onMemberRestrictionCreated = async (d: EventResponse<EventTopic.MemberRestrictionCreated>): Promise<void> => {
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        if (d.topic !== EventTopic.MemberRestrictionCreated) return;
+
```
bwpikaard, line 28: Add check for restriction type here instead of service call?

0c370t: I think it makes sense where it is, because the method is specifically named `sendQueueBanNotification`, so it has queue ban in there.

common/src/service-connectors/core/schemas/MemberRestriction.schema.ts
0c370t: Was this discussed outside of Github?

common/src/service-connectors/core/types/Member.ts
```diff
@@ -0,0 +1,17 @@
...
+});
+
+export const MemberSchema = z.object({
+    id: z.number(),
+    user: UserSchema,
+    organization: OrganizationSchema,
+});
+
+export type MemberProfile = z.infer<typeof MemberProfileSchema>;
+export type Member = z.infer<typeof MemberSchema>;
```
0c370t, line 17: Same issue as above

microservices/notification-service/src/member/member.event-subscriber.ts
```diff
@@ -0,0 +1,31 @@
...
+    constructor(
+        private readonly eventsService: EventsService,
+        private readonly memberService: MemberService,
+    ) {}
+
+    async onApplicationBootstrap(): Promise<void> {
+        await this.eventsService.subscribe(EventTopic.MemberRestrictionCreated, false).then(obs => {
+            // eslint-disable-next-line @typescript-eslint/no-misused-promises
+            obs.subscribe(this.onMemberRestrictionCreated);
```
0c370t, line 21: Why do we need eslint disables for this?

Please add `.catch(this.logger.error.bind(this.logger))` to resolve

common/src/service-connectors/core/schemas/MemberRestriction.schema.ts
bwpikaard: Yeah. As far as other microservices are aware, core models don’t exist and these are “that thing”. If they’re needed in core, it can follow the same pattern as scrims and do ISomeThing. We’ve got conflicts in core itself anyway between database and other modules.

PR #110, opened 2022-06-07
Personal/gankoji/refactor restriction messages

gankoji:
Refactor member restrictions messages. 

Previously, both restriction creation (Member banned) and restriction expiration (member unbanned) were sent as messages through the same pipe/topic in events service. To make this easier to sort out for other services, we've split those into two message streams. The pubsub, which makes its way to the front end, stays as one message stream with extra metadata as to which message type we're sending, but does need some refactoring for the type changes.


clients/web/src/lib/api/queries/RestrictedPlayers.store.ts
```diff
@@ -110,12 +109,29 @@ export class RestrictedPlayersStore extends LiveQueryStore<RestrictedPlayersStor
...
+            const memberRestriction = {
+                id: message.data.followRestrictedMembers.id,
+                type: message.data.followRestrictedMembers.type,
+                expiration: message.data.followRestrictedMembers.expiration,
+                reason: message.data.followRestrictedMembers.reason,
+                member: message.data.followRestrictedMembers.member,
+                manualExpiration: message.data.followRestrictedMembers.manualExpiration,
+                manualExpirationReason: message.data.followRestrictedMembers.manualExpirationReason,
+                memberId: message.data.followRestrictedMembers.memberId,
+            };
```
bwpikaard, line 127:
```suggestion
            const {eventType, restriction} = message.data.followRestrictedMembers;
```

clients/web/src/lib/api/queries/RestrictedPlayers.store.ts
```diff
@@ -20,32 +20,30 @@ interface Member {
...
+    manualExpirationReason?: string;
 
-        manualExpirationReason?: string;
+    memberId: number;
+}
 
-        memberId: number;
-    };
+export interface MemberRestrictionEvent extends MemberRestriction{
+    eventType: number;
```
bwpikaard, line 42: Should this be a number or enum for clarity?

bwpikaard (approved):

clients/web/src/lib/api/queries/RestrictedPlayers.store.ts
```diff
@@ -110,12 +109,29 @@ export class RestrictedPlayersStore extends LiveQueryStore<RestrictedPlayersStor
...
+                type: message.data.followRestrictedMembers.type,
+                expiration: message.data.followRestrictedMembers.expiration,
+                reason: message.data.followRestrictedMembers.reason,
+                member: message.data.followRestrictedMembers.member,
+                manualExpiration: message.data.followRestrictedMembers.manualExpiration,
+                manualExpirationReason: message.data.followRestrictedMembers.manualExpirationReason,
+                memberId: message.data.followRestrictedMembers.memberId,
+            };
+
             switch (message.data.followRestrictedMembers.eventType) {
```
bwpikaard, line 129:
```suggestion
            switch (eventType) {
```

clients/web/src/lib/api/queries/RestrictedPlayers.store.ts
bwpikaard: Actually, it's used as an enum in the switch case below?

gankoji: done

PR #119, opened 2022-06-09
Feat/services timeout

CodeRedJack:
Timeout has been added to the Service Connectors

### DIFF
--- a/common/src/global.types.ts
+++ b/common/src/global.types.ts
@@ -5,9 +5,14 @@ export enum CommonClient {
     ImageGeneration = "CommonImageGeneration",
     Notification = "CommonNotification",
     Core = "CommonClientCore",
+
 }
 
 export enum ResponseStatus {
     SUCCESS = "SUCCESS",
     ERROR = "ERROR",
+}
+
+export interface MicroserviceRequestOptions {
+    timeout?: number;
 }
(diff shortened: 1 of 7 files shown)


common/src/global.types.ts
```diff
@@ -5,9 +5,14 @@ export enum CommonClient {
     ImageGeneration = "CommonImageGeneration",
     Notification = "CommonNotification",
     Core = "CommonClientCore",
+
```
bwpikaard, line 8: Remove this line

microservices/notification-service/src/scrim/scrim.service.ts
```diff
@@ -1,4 +1,5 @@
 import {Injectable, Logger} from "@nestjs/common";
+import {MICROSERVICES_PACKAGE_NOT_FOUND_EXCEPTION} from "@nestjs/core/errors/messages";
```
bwpikaard, line 2: What does this import do? It doesn't appear to be used.

bwpikaard (approved):

microservices/notification-service/src/scrim/scrim.service.ts
CodeRedJack: It does nothing. It was an autoimport from a bad autocomplete of microserviceOptions . 
making the changes and will push them shortly

0c370t (approved):

PR #120, opened 2022-06-10
Create Sprocket Rating v0

gankoji:
This PR creates the new SprocketRating service and module. Module is defined at the `src` level in the hierarchy, so it can be depended upon by both `FinalizationService` (`ReplayParseModule`) and `MLEDBScrimService` (`MLEDBInterfaceModule`). This service calculates the new Sprocket Rating metric, as defined by the DS team. Additionally, we adjust the image-gen query template to grab this new field from the database for report cards.


core/.eslintrc.cjs
```diff
@@ -539,7 +539,7 @@ module.exports = {
     "@typescript-eslint/no-dupe-class-members": "error",                       /* (   ) Disallow duplicate class members */
     "@typescript-eslint/no-duplicate-imports": "error",                        /* (   ) Disallow duplicate imports */
     "@typescript-eslint/no-empty-function": "off",                             /* (r  ) Disallow empty functions */
-    "@typescript-eslint/no-extra-parens": "error",                             /* ( f ) Disallow unnecessary parentheses */
+    "@typescript-eslint/no-extra-parens": "off",                               /* ( f ) Disallow unnecessary parentheses */
```
bwpikaard, line 542: Can this be disabled on the lines where it needs to be instead of globally?

core/src/app.module.ts
```diff
@@ -16,6 +16,7 @@ import {OrganizationModule} from "./organization";
 import {ReplayParseModule} from "./replay-parse";
 import {SchedulingModule} from "./scheduling";
 import {ScrimModule} from "./scrim";
+import {SprocketRatingModule} from "./sprocket-rating/sprocket-rating.module";
```
bwpikaard, line 19: Barrel file

bwpikaard (approved):

core/.eslintrc.cjs
gankoji: We set it to off because we straight up don't ever want it.

core/src/database/mledb/PlayerStatsCore.model.ts
```diff
@@ -56,12 +56,27 @@ export class MLE_PlayerStatsCore {
     @Column("integer", {name: "assists"})
     assists: number;
 
+    @Column("integer", {name: "goals_against"})
+    goals_against: number;
+
+    @Column("integer", {name: "shots_against"})
+    shots_against: number;
+
```
0c370t, line 64: Need to be nullable

core/src/sprocket-rating/sprocket-rating.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    opi: number;
+    dpi: number;
+    gpi: number;
+}
+
+@Injectable()
+export class SprocketRatingService {
+    private logger = new Logger(SprocketRatingService.name);
+
+    calcSprocketRating(core: Partial<MLE_PlayerStatsCore>): SprocketRating {
```
0c370t, line 15: Not sure how much this would change things; but can we break the dependence on the MLEDB type for this?

core/src/sprocket-rating/sprocket-rating.service.ts
```diff
@@ -0,0 +1,61 @@
...
+            const dpi = 100.0 / (1 + Math.exp(DPI_beta * ((dpi_raw - DPI_mu) / DPI_sigma)));
+
+            const gpi = (opi + dpi) / 2.0;
+
+            return {
+                opi, dpi, gpi,
+            };
+        }
+
+        this.logger.warn(`Received null data in sprocket rating calculation. ${core.goals}, ${core.assists}, ${core.shots}, ${core.saves}, ${core.goals_against}, ${core.shots_against}.`);
```
0c370t, line 53: Should we default to 0 in these cases?

0c370t (changes requested):

core/src/database/mledb/PlayerStatsCore.model.ts
gankoji: I don't see why these need to be nullable but the other core stats don't.

core/src/sprocket-rating/sprocket-rating.service.ts
gankoji: Seems like the sensible option here, easy to check for and bug out of the calculation. If there's a better default value we can change to that.

core/src/sprocket-rating/sprocket-rating.service.ts
gankoji: Really it would either drastically change the API for this method (by passing in each field separately) or mandate yet another payload type specific to this calculation. We can probably think that through a bit better down the line.

core/src/database/mledb/PlayerStatsCore.model.ts
0c370t: It doesn't exist in every column that already exists, so we either need a default or a nullable, nullable would make more sense to me because we don't have the legacy data

core/src/sprocket-rating/sprocket-rating.service.ts
gankoji: Type added.

core/src/database/mledb/PlayerStatsCore.model.ts
gankoji: Done.

PR #134, opened 2022-06-17
Feat: Click Outside Close 

CodeRedJack:
clickOutsideAction.ts simplified. Will need to reintroduce the code later to reduce overhead in some instances where clickOutside may be implemented. 

Also branch name is meh, I made this for one ticket and jumped to another.

### DIFF
--- a/clients/web/src/lib/components/atoms/Modal.svelte
+++ b/clients/web/src/lib/components/atoms/Modal.svelte
@@ -14,7 +14,9 @@
     };
 
     const handleClickOutside = (): void => {
+        console.log("Handler fired.");
         visible = false;
+    }
 
     setContext("close", close);
 </script>
@@ -22,8 +24,8 @@
 <Portal>
     <input type="checkbox" class="modal-toggle" bind:checked={visible}/>
 
-    <div class="modal bg-gray-700/40" {id} use:clickOutside on:outclick = "{ handleClickOutside() }">
-        <div class="modal-box max-w-xl">
+    <div class="modal bg-gray-700/40" {id} >
+        <div class="modal-box max-w-xl" visible={visible} use:clickOutside on:outclick={handleClickOutside}>
             <div class="close" on:click={close}>
                 <IoMdClose/>
             </div>
--- a/clients/web/src/lib/components/atoms/Modal.svelte
+++ b/clients/web/src/lib/components/atoms/Modal.svelte
@@ -7,19 +7,14 @@
 
     export let title: string;
     export let visible: boolean = false;
-    export let canClickOutside: boolean = false;
+    export let canClickOutside: boolean = true;
     export let id: string;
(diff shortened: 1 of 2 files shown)


gankoji (approved):

clients/web/src/lib/utils/actions/clickOutsideAction.ts
```diff
@@ -1,34 +1,13 @@
...
-    
-    addListener: function(func: CallableFunction, el: HTMLElement) {
-        if (this.registeredListeners.size === 0) {
-            // Start listening to document clicks
-            this.boundFunction = this.handleOnClick.bind(this);
-            document.addEventListener("click", this.boundFunction);
+export function clickOutside(node) {
+    const handleClick = event => {
+        if (!node.contains(event.target)) {
+            node.dispatchEvent(new CustomEvent("outclick"));
```
0c370t, line 4: Why are custom events required here?

clients/web/src/lib/components/atoms/Modal.svelte
```diff
@@ -7,36 +7,36 @@
 
     export let title: string;
     export let visible: boolean = false;
-    export let canClickOutside: boolean = false;
```
0c370t, line 10: Why remove this prop?

CodeRedJack: It wasn't in use. So I thought it was better coding to clear it out. It was showing grey in the IDE. I was thinking like an unused import. 

The only time handle clickoutside can be called is if you are clicking outside the modal. Then you don't need to track they in that file. You just set visible to false and it seemed to run the remove part of the click outside. 

It was how the svelte tutorial did it but I am probably missing something 😅

clients/web/src/lib/utils/actions/clickOutsideAction.ts
CodeRedJack: Because the modal uses on:outclick. The customs event is distinguished from a click on the buttons of the modals this way. (At least I think this is how that works) a click is a click otherwise right?

clients/web/src/lib/components/atoms/Modal.svelte
0c370t: This prop was in place such that a modal might not let you click outside to close (i.e. it may have some required stuff a user must do)

0c370t: We should restore it

clients/web/src/lib/utils/actions/clickOutsideAction.ts
0c370t: You can distinguish where a click happens by using el.contains, similarly to the previous implementation; I don't think we should add a custom event that fires for this

CodeRedJack:
Fixing in another branch

PR #135, opened 2022-06-17
Lock Scrim After 3 Rejections

bwpikaard:


microservices/matchmaking-service/src/scrim/scrim.event-subscriber.ts
```diff
@@ -100,6 +106,11 @@ export class ScrimEventSubscriber {
             this.logger.warn(`Scrim not found for submission ${payload.submissionId}`);
             return;
         }
-        await this.scrimService.resetScrim(scrim.id);
+        await this.scrimService.resetScrim(scrim.id).catch(() => null);
```
bwpikaard, line 109: Caused instances of "scrim was already in x state" without catching. @0c370t look into?

0c370t: Not sure where that specific error would come from; nothing in this function does that, except for the check that the scrim is in ratifying; which _should_ be the case at this point. Not a huge fan of just swallowing the error for the sake of swallowing it.

microservices/submission-service/src/replay-submission/replay-submission-crud/replay-submission-crud.controller.ts
```diff
@@ -15,4 +15,11 @@ export class ReplaySubmissionCrudController {
         await this.crudService.removeSubmission(data.submissionId);
         return {removed: true};
     }
+
+    @MessagePattern(SubmissionEndpoint.GetSubmissionRejections)
+    async getSubmissionRejections(@Payload() payload: unknown): Promise<IGetSubmissionRejections_Response> {
```
0c370t, line 20: I think there is a way to do this, without needing to create an interface; but I could be wrong. Lets look into it on Sunday.

0c370t (changes requested):

microservices/matchmaking-service/src/scrim/scrim.event-subscriber.ts
bwpikaard: That seem better?

microservices/submission-service/src/replay-submission/replay-submission-crud/replay-submission-crud.controller.ts
0c370t: `SubmissionOutput<SubmissionEndpoint.GetSubmissionRejections>`

PR #148, opened 2022-06-26
fix: grab image-gen-frontend dev branch

andreasmwenzel:
This should fix image gen frontend.

All I did was replace the src/ folder and added some of the configs from image-generation-frontend's dev branch


andreasmwenzel:
Git Guardian doesn't like something in this. I think it might be related to the ca-certificate?

PR #156, opened 2022-07-10
Scrim report cards 2

andreasmwenzel:


core/src/image-generation/image-generation.service.ts
```diff
@@ -23,8 +23,8 @@ export class ImageGenerationService {
         const result = await this.igService.send(
             ImageGenerationEndpoint.GenerateImage,
             {
-                inputFile: `scrim_report_cards/scrimReportCards/template.svg`,
-                outputFile: `scrim_report_cards/scrimReportCards/outputs/${scrimId}_1`,
+                inputFile: `scrim_report_cards/scrimReportCards2/template.svg`,
+                outputFile: `scrim_report_cards/scrimReportCards2/outputs/${scrimId}_1`,
```
andreasmwenzel, line 27: This is the part that actually points core to the new files. The queries are already updated on the db but do not affect current scrim report cards

gankoji (approved):

0c370t (approved):

PR #181, opened 2022-08-12
Fix Scrim Timeouts

bwpikaard:


gankoji (approved):

microservices/matchmaking-service/src/scrim/scrim.module.ts
```diff
@@ -31,6 +31,7 @@ import {ScrimMetricsService} from "./scrim-metrics/scrim-metrics.service";
                             servername: config.redis.host,
                         }
                     : undefined,
+                keyPrefix: `${config.redis.prefix}:bull`,
```
0c370t, line 34: Could this value come from the config?

PR #177, opened 2022-07-31
CU-2gwd894 Submission Ready for Ratifications Notifications

bwpikaard:

PR #182, opened 2022-08-12
Match Report Cards CU-2t9fgma

bwpikaard:


0c370t:
Task linked: [CU-2t9fgma Wire up League Submissions to generate Report Cards](https://app.clickup.com/t/2t9fgma)

bwpikaard:
Depends on #177

bwpikaard:
Waiting on #195

gankoji (approved):
One thought: looks like report card generation is bounded by a 300s (5 min) timeout currently. I know we expect report cards to be expensive: are we worried at all about getting too many matches or scrims or both submitted at once and overloading our system? Async methods will still stall at `await`, and we only have so many threads. Something to consider keeping an eye on once we go live to the full audience.

0c370t:
Regarding this, in theory everything is horizontally scalable. I have a plan to turn everything to run 2 instances in staging and dev; which should give us a smoke test on parallelization, and we can scale the services in production as needed.

The biggest things I see causing issues here are replay parsing and image generation; so we might go to some crazy number like 5-10 instances, especially given that we are running on "free" compute

PR #189, opened 2022-08-19
Fix Matchup Checking CU-3t4vjan

bwpikaard:


0c370t:
Task linked: [CU-3t4vjan Make Game Order not Matter](https://app.clickup.com/t/3t4vjan)

PR #208, opened 2022-09-01
Add 'AddNewPlayer' admin/LO action for ELO, and pull data from MLEDB to populate ELODB

gankoji:


core/src/elo/elo.service.ts
```diff
@@ -261,4 +262,54 @@ export class EloService {
...
+                return 3;
+            }
+            case "MASTER": {
+                return 4;
+            }
+            case "PREMIER": {
+                return 5;
+            }
+            default: {
+                return 6;
```
bwpikaard, line 311: Throw new error here because if it doesn't exist we don't want random things happening

core/src/elo/elo-connector/schemas/AddPlayer.schema.ts
```diff
@@ -0,0 +1,24 @@
+import {z} from "zod";
+
+export enum SkillGroups {
+    FOUNDATION = 1,
+    ACADEMY = 2,
+    CHAMPION = 3,
+    MASTER = 4,
+    PREMIER = 5,
+    NONE = 6,
+}
```
bwpikaard, line 10:
```suggestion
export enum SkillGroup {
    FOUNDATION = 0,
    ACADEMY = 1,
    CHAMPION = 2,
    MASTER = 3,
    PREMIER = 4,
}
```

core/src/elo/elo-connector/schemas/AddPlayer.schema.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const NewPlayerSchema = z.object({
+    id: z.number(),
+    name: z.string(),
+    salary: z.number(),
+    skillGroup: z.nativeEnum(SkillGroups),
```
bwpikaard, line 16:
```suggestion
    skillGroup: z.nativeEnum(SkillGroup),
```

core/src/elo/elo-connector/schemas/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from "./AddPlayer.schema";
```
bwpikaard, line 1: Make name consistent with what is in the file.
```suggestion
export * from "./AddPlayers.schema";
```

core/src/elo/elo.resolver.ts
```diff
@@ -0,0 +1,25 @@
...
+    @Mutation(() => String)
+    async generateMigrationData(): Promise<string> {
+        await this.eloService.prepMigrationData();
+        return "Done.";
+    }
+
+    @Mutation(() => String)
+    async runEloMigration(): Promise<string> {
+        const inputData = await this.eloService.prepMigrationData();
+        await this.eloConnectorService.createJob<EloEndpoint.AddNewPlayers>(EloEndpoint.AddNewPlayers, inputData);
```
bwpikaard, line 22: Should need generic type here; it comes from the enum parameter automatically. If not, that may be a fix on my branch.

bwpikaard (changes requested):

core/src/elo/elo.service.ts
gankoji: Leaving this open for now, as there do exist players with skillGroup "UNKNOWN" in the mledb schema (and thus in the view that we're pulling this data from). TODO is in the code to review this once the MLEDB->Sprocket migration is complete, and the view is updated to pull from players in the sprocket schema.

bwpikaard (changes requested):
Need shuckle to rework the view to base off of Sprocket players rather than MLEDB players since that's what elo will be based off of. Otherwise, looks good.

core/src/elo/elo.service.ts
```diff
@@ -261,4 +262,57 @@ export class EloService {
         const seriesTypeStr = series.matchParent.fixture ? "fixture" : series.matchParent.scrimMeta ? "scrim" : "unknown";
         return `\`seriesId=${seriesId}\` ${seriesTypeStr ? `(${seriesTypeStr})` : ""} successfully marked \`fullNcp=${isNcp}\` with updated elo, and all connected replays had their elo updated.${numReplays && dummiesNeeded ? ` **${dummiesNeeded} dummy replay(s)** were added to the series.` : ""}`;
     }
+
+    async prepMigrationData(): Promise<NewPlayer[]> {
+        // Run queries, first refreshing the materialized view
+        this.logger.verbose("Refreshing materialized view.");
+        // await this.dataSource.manager.query("REFRESH MATERIALIZED VIEW mledb.v_current_elo_values");
```
zachstence, line 269: Is this supposed to be commented out? If we don't need it, remove this and log above

core/src/elo/elo.service.ts
```diff
@@ -261,4 +262,57 @@ export class EloService {
...
+
+        // Then querying from it
+        /* eslint-disable @typescript-eslint/no-unsafe-assignment,
+        @typescript-eslint/no-explicit-any */
+        this.logger.verbose("Querying the materialized view.");
+        const rawData: any[] = await this.dataSource.manager.query("SELECT player_id, elo, league, salary, name FROM mledb.v_current_elo_values");
```
zachstence, line 275: You should write an interface for what this `rawData` looks like to avoid using `any`

zachstence (changes requested):
I don't have the full picture of ELO or this DB stuff, but I have a few comments regarding JS

core/src/elo/elo.service.ts
gankoji: This is needed, I have uncommented.

core/src/elo/elo.service.ts
gankoji: Fixed

bwpikaard (approved):

PR #210, opened 2022-09-01
Sprocket event marshal [CU-3u734kc]

bwpikaard:


0c370t:
Task linked: [CU-3u734kc Notification Service Decorator Based Events](https://app.clickup.com/t/3u734kc)

common/src/events/marshal/marshal.types.ts
```diff
@@ -0,0 +1,13 @@
+import {z} from "zod";
+
+import type {EventPayload} from "../events.types";
+import {EventTopic} from "../events.types";
+
+export type EventFunction<Event extends EventTopic> = (data: EventPayload<Event>) => Promise<void>;
```
zachstence, line 6: Not all event functions will be `async`

```suggestion
export type EventFunction<Event extends EventTopic> = (data: EventPayload<Event>) => void | Promise<void>;
```

zachstence (changes requested):
This is awesome, can't wait to use this more throughout the code

🚀

common/src/events/marshal/marshal.ts
```diff
@@ -39,7 +39,13 @@ export abstract class SprocketEventMarshal {
                         return;
                     }
 
-                    eventFunction(payload.data).catch(e => { this.logger.error(e) });
+                    try {
+                        // No need to catch; it's in a try-catch block because we don't know if it's a promise or not.
+                        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+                        eventFunction(payload.data);
```
zachstence, line 45: If its async, you have to `await` the Promise in order for the error to be caught in the try/catch.

You should be able to use `instanceof` to determine if its a Promise or not

```ts
try {
    const maybePromise = someFunction();
    if (maybePromise instanceof Promise) {
        await maybePromise;
    }
} catch (e) {
    // ...
}
```

zachstence (changes requested):

zachstence (approved):

PR #223, opened 2022-09-10
fix: return undefined when subscribing to a scrim, but the user isnt in one

zachstence:


0c370t:
Task linked: [CU-2xjq8ae Prevent Error from being thrown when subscribing to currentScrim](https://app.clickup.com/t/2xjq8ae)

PR #227, opened 2022-09-10
Personal/gankoji/lo rankouts

gankoji:


bwpikaard (approved):

core/src/franchise/player/player.resolver.ts
```diff
@@ -54,4 +77,127 @@ export class PlayerResolver {
...
+            },
+            new: {
+                id: skillGroup.id,
+                name: skillGroup.profile.description,
+                salary: Number(salary),
+                discordEmojiId: skillGroup.profile.discordEmojiId,
+            },
+        });
+
+        await this.notificationService.send(NotificationEndpoint.SendNotification, {
```
0c370t, line 155: Should this be handled whenever a rank out occurs?

core/src/franchise/player/player.resolver.ts
```diff
@@ -54,4 +77,127 @@ export class PlayerResolver {
...
+            where: {
+                user: {
+                    id: player.member.user.id,
+                },
+                accountType: UserAuthenticationAccountType.DISCORD,
+            },
+        });
+        const orgProfile = await this.organizationService.getOrganizationProfileForOrganization(player.member.organization.id);
+
+        if (player.skillGroup.id === skillGroupId) return "ERROR: You are already in this skill group";
```
0c370t, line 124: This player is already

core/src/franchise/player/player.resolver.ts
gankoji: We currently get DM notifications from emilio whenever LO takes direct action, like for a manual rankout. I don't see why we wouldn't continue with that behavior.

PR #235, opened 2022-09-12
fix: move player directly to skill group input

gankoji:


bwpikaard (approved):

core/src/franchise/player/player.service.ts
```diff
@@ -416,6 +416,41 @@ export class PlayerService {
...
+
+        let player = await this.mle_playerRepository.findOneOrFail({
+            where: {
+                discordId: discId.accountId,
+            },
+        });
+
+        player = this.mle_playerRepository.merge(player, {
+            role: Role.NONE,
+            teamName: "WW",
```
zachstence, line 444: Team name should only change on rankdown

zachstence (changes requested):

bwpikaard (approved):

PR #247, opened 2022-09-14
fix: use name from commandSpec

zachstence:
Use `name` from command decorator so that it doesn't make a difference if a user calls the command with different capitalization


0c370t:
Task linked: [CU-2y9q2mb Analytics, ensure all strings are lower cased](https://app.clickup.com/t/2y9q2mb)

PR #258, opened 2022-09-16
Webhooks

bwpikaard:


core/migrations/1663308333261-Automigration.ts
```diff
@@ -0,0 +1,22 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Automigration1663308333261 implements MigrationInterface {
+    name = 'Automigration1663308333261'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE TYPE "sprocket"."webhook_type_enum" AS ENUM('ScrimReportCard', 'MatchReportCard', 'SubmissionReadyForRatification', 'ScrimCreated')`);
```
0c370t, line 7: We don't want this; instead we should use Foreign Keys to relate things like this.

i.e. Franchise Profile would have `matchWebhookId` / `scrimWebhookId`

bwpikaard: What's the point of a webhook table if it'll only contain a webhook url? What's the difference?

PR #271, opened 2022-09-17
Implement refresh tokens for authentication

gankoji:
Old behavior:

- Log in
- Get JWT from the backend, valid for 7 days from the first login. 
- After 7 days, you have to log in again, regardless of last usage.

New behavior:

- Log in
- Get JWT from the backend, valid for 30 minutes, as well as a refresh token valid for 7 days.
- If at any point after 30 minutes but before 7 days you refresh the app, sprocket will *automatically* refresh your tokens, and you'll be logged in for another 30 minutes (with a new refresh token, valid for 7 days from the *refresh point*)

Thus, the only way a user will have to start the authorization process from scratch is if they haven't used Sprocket in more than 7 days.

### DIFF
--- a/clients/web/src/hooks.ts
+++ b/clients/web/src/hooks.ts
@@ -2,23 +2,83 @@ import type {GetSession, Handle} from "@sveltejs/kit";
 import {constants, loadConfig} from "$lib/utils";
 
 export const handle: Handle = async ({event, resolve}) => {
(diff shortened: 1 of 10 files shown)


bwpikaard (approved):
LGTM

0c370t (approved):

PR #279, opened 2022-09-19
Reprocess Matches

gankoji:
For data integrity purposes (particularly wrt to Elo), we want to have the ability to "hard reset" and reprocess all match data available from a given date/time. This command allows for the second half of that process (the first, deleting the desired elo data, must still be done manually).

### DIFF
--- a/core/src/scheduling/match/match.service.ts
+++ b/core/src/scheduling/match/match.service.ts
@@ -5,7 +5,7 @@ import type {
 } from "@sprocketbot/common";
 import type {FindOneOptions, FindOptionsRelations} from "typeorm";
 import {
-    IsNull, Not, Repository,
+    IsNull, MoreThan, Not, Repository,
 } from "typeorm";
 
 import type {
@@ -16,6 +16,7 @@ import {
     Franchise,
     Invalidation,
     Match,
+    MatchParent,
     PlayerStatLineStatsSchema,
     Round,
     ScheduleFixture,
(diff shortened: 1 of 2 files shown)


core/src/scheduling/match/match.service.ts
```diff
@@ -115,6 +117,37 @@ export class MatchService {
...
+
+        const jobs = Promise.all(matchParents.map(async mp => {
+            if (mp.event) return;
+            let isScrim = false;
+            if (mp.fixture) {
+                isScrim = true;
+            }
+
+            const payload = await this.translatePayload(mp.match.id, isScrim);
+            await this.eloConnectorService.createJob(EloEndpoint.CalculateEloForMatch, payload);
```
bwpikaard, line 145: We want to do this chronologically somehow. Is there any way we can force this? We'd want to do this synchronously, not a Promise.all.

bwpikaard (changes requested):

core/src/scheduling/match/match.service.ts
gankoji: You're right, I'd forgotten about that. I'll have to do an order by and then a synchronous for loop.

core/src/scheduling/match/match.service.ts
```diff
@@ -115,6 +116,31 @@ export class MatchService {
...
+                                   mp."fixtureId" IS NULL AND mp."scrimMetaId" IS NULL     AS broken
+                                FROM round_played_time
+                                         INNER JOIN match m ON "matchId" = m.id
+                                         INNER JOIN match_parent mp ON m."matchParentId" = mp.id
+                                WHERE played_at > :start_date
+                                GROUP BY "matchId", mp.id, m.id
+                                ORDER BY 2;`;
+
+        interface toBeReprocessed {id: number; date: string; is_league_match: boolean;}
+        const results: toBeReprocessed[] = await this.dataSource.manager.query(queryString, [ {start_date: startDate} ]) as toBeReprocessed[];
```
0c370t, line 136:
```suggestion
                                WHERE played_at > $1
                                GROUP BY "matchId", mp.id, m.id
                                ORDER BY 2;`;

        interface toBeReprocessed {id: number; date: string; is_league_match: boolean;}
        const results: toBeReprocessed[] = await this.dataSource.manager.query(queryString, [startDate]) as toBeReprocessed[];
```

bwpikaard (approved):

PR #299, opened 2022-09-25
Fix Salaries and Rankouts

bwpikaard:


core/src/database/franchise/player/player.model.ts
```diff
@@ -26,7 +26,13 @@ export class Player extends BaseModel {
...
     skillGroupId: number;
 
-    @Column({type: "numeric"})
+    @Column({
+        type: "numeric",
+        transformer: {
+            from: (value: string) => parseFloat(value),
+            to: (value: number) => value.toString(),
+        },
+    })
```
zachstence, line 35: Why is this necessary again? Why don't we need this on other number fields?

zachstence (changes requested):

core/src/database/franchise/player/player.model.ts
bwpikaard: `numeric` is returned from typeorm as a string.

zachstence: Why don't we change the column type in the DB?

bwpikaard: Not sure. That's just what was there. I was going to change it to `double` or `float` but wasn't sure if there was a reason.

gankoji (dismissed):

core/src/database/franchise/player/player.model.ts
zachstence: Lets change the column type

zachstence (approved):

PR #308, opened 2022-09-27
Fix FA Intake

bwpikaard:


core/src/franchise/player/player.service.ts
```diff
@@ -170,6 +170,9 @@ export class PlayerService {
         timezone: Timezone,
         modePreference: ModePreference,
     ): Promise<Player> {
+        const mleOrg = await this.organizationRepository.findOneOrFail({where: {profile: {name: "Minor League Esports"} } });
```
0c370t, line 173: Relation is likely missing

bwpikaard:
```suggestion
        const mleOrg = await this.organizationRepository.findOneOrFail({where: {profile: {name: "Minor League Esports"} }, relations: {profile: true} });
```

gankoji (approved):

PR #310, opened 2022-09-30
Fix: Show admin tab without refresh

zachstence:
The admin tab was getting added to the navigation context in a `load` function that read the user from the session.

This didn't work because
1. The session doesn't contain the user in `load` during SSR
2. Context doesn't update reactively

Moving it to a store and checking for admin on page mount fixes the problem.


0c370t:
Task linked: [CU-3ygefp0 Admin Tab Bug](https://app.clickup.com/t/3ygefp0)

0c370t (approved):

PR #328, opened 2022-10-08
Fix Refresh Tokens

gankoji:
I have intentionally left the expiry time on the JWT very short (1 minute) so that this is easy to test in staging. Once we're satisfied that this works properly for everyone I will cut another PR to change that expiry to a value we agree is more sensible.

### DIFF
--- a/clients/web/src/hooks.ts
+++ b/clients/web/src/hooks.ts
@@ -1,15 +1,94 @@
 import type {GetSession, Handle} from "@sveltejs/kit";
 import {constants, loadConfig} from "$lib/utils";
+import {add} from "date-fns";
+
+interface refreshPayload {
+    cookies: string[];
+    cookiesString: string;
+    accessToken: string;
+    refreshToken: string;
+}
+
+const doAuthRefresh = async (refreshUrl: string, currentCookies: string): Promise<refreshPayload> => {
+
+    let newCookiesString = "";
+    // If  refresh it
+    // Get the refresh token out of user's cookies
+    const refreshToken = currentCookies.split("; ").find(c => c.split("=")[0] === constants.refresh_token_cookie_key)
+        ?.split("=")[1];
+    if (refreshToken) {
+        // Send that refresh token to the backend, asking
+        // for a new JWT
+        const res = await fetch(refreshUrl, {
+            method: "GET",
+            headers: {
+                "Authorization": `Bearer ${refreshToken}`,
+            },
+        });
+
+        // Pull the new tokens out of the response
+        const tokens = await res.json();
+        const access_token = tokens.access_token;
+        const new_refresh_token = tokens.refresh_token;
+
+        // Store these new tokens back in the user's cookies
+        const newCookies = currentCookies.split("; ");
+
+        // Access token cookie
+        let newCookies1: string[];
+        if (newCookies.some(c => c.split("=")[0] === constants.auth_cookie_key)) {
+            // Cookie still exists, just update it
+            newCookies1 = newCookies.map(c => {
+                if (c.split("=")[0] === constants.auth_cookie_key) {
+                    return `${constants.auth_cookie_key}=${access_token};expires=${add(new Date(), {weeks: 1}).toUTCString()}`;
+                }
+                return c;
+            });
+        } else {
+            // Access token cookie doesn't exist, add it.
+            newCookies1 = newCookies;
+            newCookies1.push(`${constants.auth_cookie_key}=${access_token};expires=${add(new Date(), {weeks: 1}).toUTCString()}`);
+        }
+
+        // Refresh token cookie
+        const newCookies2 = newCookies1.map(c => {
+            if (c.split("=")[0] === constants.refresh_token_cookie_key) {
+                return `${constants.refresh_token_cookie_key}=${new_refresh_token};expires=${add(new Date(), {weeks: 1}).toUTCString()}`;
+            }
+            return c;
+        });
+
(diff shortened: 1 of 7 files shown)


clients/web/src/routes/auth/login.svelte
```diff
@@ -26,7 +26,9 @@
         $session.user = extractJwt<SessionUser>(tokens[0]);
         $session.token = tokens[0];
 
-        cookies.set(constants.auth_cookie_key, tokens[0], {expires: 0.25}); // 6 hours
+	// It is important that the auth cookie lasts as long as the refresh token. Do not shorten it,
```
bwpikaard, line 29: I don't really agree with that but alright. If the jwt expires, we should be checking that a refresh token exists when a regular token doesn't, not having a stale/expired auth token.

gankoji: I forgot to mention this in my PR comments, but I'm ok with refactoring that behavior to incorporate what you had done with the draft PR. I just set things up this way for minimal changes while debugging/testing.

bwpikaard: I don't particularly care either way, just don't like the feeling of keeping the stale token around. If it works, it works though.

core/src/identity/auth/oauth/oauth.controller.ts
```diff
@@ -54,9 +56,10 @@ export class OauthController {
     @UseGuards(JwtRefreshGuard)
     @Get("refresh")
     async refreshTokens(@Request() req: Req): Promise<AccessToken> {
```
zachstence, line 58: How is the token actually refreshed? I.e., when/where is this `/refresh` endpoint called from the client? Shouldn't every request the client makes be able to refresh the token if it is expired?

clients/web/src/routes/auth/login.svelte
zachstence: Yeah I don't see why we would keep the expired token around for this long 🤔

core/src/identity/auth/oauth/oauth.controller.ts
gankoji: https://github.com/SprocketBot/sprocket/blob/8a89082050029cef3bce6217ff47e416352cc426/clients/web/src/hooks.ts#L31 is where it's called right now, from hooks.ts

gankoji: I have seen solutions that use interceptors with Axios to do what you're describing (refresh on any request) as well. This solution just uses what we currently have.

clients/web/src/routes/auth/login.svelte
gankoji: It's only for the purpose of the check (is token expired) in hooks. Can't check a token's expiry if the cookie it lives in is deleted (due to being expired itself). Hyper has a draft PR up that inverts that check and makes this OBE, but we just haven't integrated it yet. I can do that if we'd prefer it.

gankoji: Changed, see latest commit

gankoji:
Added https://app.clickup.com/t/3nuzbuq to the backlog to address refreshing on any request to the backend. That's going to be a good bit more effort.

clients/web/src/hooks.ts
```diff
@@ -1,15 +1,94 @@
...
+            accessToken: access_token,
+            refreshToken: new_refresh_token,
+        };
+    }
+
+    return {
+        cookies: [""],
+        cookiesString: "",
+        accessToken: "",
+        refreshToken: "",
```
0c370t, line 76: Do we know if this will wipe a user's cookies? If so, is that the behavior that we want?

core/src/identity/auth/oauth/oauth.service.ts
```diff
@@ -19,7 +19,7 @@ export class OauthService {
 
     async loginDiscord(user: AuthPayload): Promise<AccessToken> {
         return {
-            access_token: this.jwtService.sign(user, {expiresIn: "6h"}),
+            access_token: this.jwtService.sign(user, {expiresIn: "1m"}),
```
0c370t, line 22: This seems like a great spot for a config file value

gankoji: I have a ticket in clickup for exactly this, but we deemed it a pretty low priority. I can always sneak that in here.

clients/web/src/hooks.ts
gankoji: I don't think so, but I haven't tested it. Sending an empty "Set-cookies" HTTP request shouldn't wipe cookies, just not add any new ones.

core/src/identity/auth/oauth/oauth.service.ts
gankoji: Fixed in latest commit

core/config/default.json
```diff
@@ -49,6 +49,8 @@
       "callbackUrl": ""
     },
     "jwt_expiry": "",
+    "access_expiry": "",
+    "refresh_expiry": "",
```
bwpikaard, line 53: Set a default value. 6h and 7d or something.

bwpikaard, line 53:
```suggestion
    "access_expiry": "6h",
    "refresh_expiry": "7d",
```

bwpikaard (changes requested):

0c370t (approved):

bwpikaard (approved):

PR #330, opened 2022-10-09
feat: upsert fa intake

bwpikaard:


gankoji (approved):

0c370t (approved):

core/src/franchise/player/player.service.ts
```diff
@@ -180,61 +180,94 @@ export class PlayerService {
         let player: Player;
 
         try {
-            const user = this.userRepository.create({});
+            const mlePlayer = await this.mle_playerRepository.findOne({where: {mleid} });
+
+            if (mlePlayer)  {
+                const bridge = await this.ptpRepo.findOneOrFail({where: {mledPlayerId: mlePlayer.id} });
```
zachstence, line 186: Why would `mlePlayer` exist if this function is to intake new members?

zachstence (changes requested):

core/src/franchise/player/player.service.ts
bwpikaard: It's an upsert.

zachstence (approved):

PR #332, opened 2022-10-11
fix: incorrect names unreported accounts

bwpikaard:


gankoji (dismissed):

core/src/franchise/player/player.controller.ts
```diff
@@ -154,4 +166,77 @@ export class PlayerController {
 
         return "SUCCESS";
     }
+
+    @MessagePattern(CoreEndpoint.GetPlayerByPlatformId)
+    async getPlayerByPlatformId(@Payload() payload: unknown): Promise<CoreOutput<CoreEndpoint.GetPlayerByPlatformId>> {
+        const data = CoreSchemas[CoreEndpoint.GetPlayerByPlatformId].input.parse(payload);
+
+        const game = await this.gameService.getGameById(data.gameId);
```
zachstence, line 174: We're only using `game.id`, which we have on `data.gameId`. Fetching `game` is unnecessary

core/src/franchise/player/player.controller.ts
```diff
@@ -154,4 +166,77 @@ export class PlayerController {
...
+        const platform = await this.platformService.getPlatformByCode(data.platform);
+        const player = await this.playerService.getPlayerByGameAndPlatform(game.id, platform.id, data.platformId).catch(() => null);
+
+        if (!player) return {
+            success: false,
+            request: data,
+        };
```
zachstence, line 181: Rather than `.catch(() => null)`, then checking for null, just put the `getPlayerByGameAndPlatform` call into a `try`/`catch`

We should also be looking for a specific error (player not found), because if its a different unexpected error we should probably halt execution. This may take some more refactoring though and not be doable in this PR.

core/src/franchise/player/player.controller.ts
```diff
@@ -154,4 +166,77 @@ export class PlayerController {
...
+                data: {
+                    id: player.id,
+                    userId: player.member.user.id,
+                    skillGroupId: player.skillGroupId,
+                    franchise: {
+                        name: mlePlayer.teamName,
+                    },
+                },
+                request: p,
+            };
```
zachstence, line 227: Duplicated code from `getPlayerByPlatformId`, this could be a function in `playerService`

microservices/submission-service/src/replay-validation/replay-validation.service.ts
```diff
@@ -109,80 +106,61 @@ export class ReplayValidationService {
...
+            };
+        }
+
         // Get platformIds from stats
-        const ballchasingIds = stats.flatMap(s => [s.blue, s.orange].flatMap(t => t.players.flatMap(p => ({
-            platform: p.id.platform,
-            id: p.id.id,
+        // Don't send the same request for multiple players
+        const uniqueBallchasingPlayerIds = Array.from(new Set(stats.flatMap(s => [s.blue, s.orange].flatMap(t => t.players.flatMap(p => ({
             name: p.name,
```
zachstence, line 121: `name` is unused, can we remove it?

zachstence (changes requested):

microservices/submission-service/src/replay-validation/replay-validation.service.ts
bwpikaard: It is used later on to return a list of invalid players. 143

zachstence (approved):

PR #333, opened 2022-10-11
`Input` form atom

zachstence:
image


clients/web/package.json
```diff
@@ -29,6 +30,7 @@
     "histoire": "^0.11.0",
     "postcss": "^8.4.14",
     "postcss-load-config": "^4.0.1",
+    "shortid": "^2.2.16",
```
0c370t, line 33: Can we use nanoid to match what's used elsewhere?

clients/web/src/lib/components/atoms/form/Input/Input.svelte
```diff
@@ -0,0 +1,129 @@
...
+    export let placeholder: string | undefined = undefined;
+    export let disabled: boolean = false;
+    export let state: FormControlState = "none";
+    export let error: string | undefined = undefined;
+
+    const labelId = `input_${shortid.generate()}`;
+    const errorId = `input-error_${shortid.generate()}`;
+</script>
+
+<div class="size-{size} state-{state}">
```
0c370t, line 22: If size and state are missing; do we want to continue attaching these classes?

clients/web/src/lib/components/atoms/form/Input/Input.svelte
```diff
@@ -0,0 +1,129 @@
...
+    label {
+        @apply block mb-2 text-sm font-medium text-gray-900 dark:text-gray-300;
+    }
+
+    .input-container {
+        @apply flex items-stretch
+            rounded-lg border overflow-hidden
+            focus-within:ring-1 focus-within:ring-primary focus-within:border-primary
+            bg-gray-50 border-gray-300 text-gray-900 placeholder-gray-600
+            dark:bg-gray-700 dark:border-gray-600 dark:text-white dark:placeholder-gray-400;
```
0c370t, line 57: We are dark mode by default; without support for light mode, so we should swap this to remove the light mode values

clients/web/src/lib/components/atoms/form/Input/Input.svelte
```diff
@@ -0,0 +1,129 @@
+<script lang="ts" context="module">
+    export type InputSize = "sm" | "md" | "lg";
```
0c370t, line 2: Can this be consolidated into the form types?

clients/web/src/lib/components/atoms/form/Input/Input.svelte
zachstence: Size and state have default values of `md` and `none` respectively, so they won't be missing, and we do want to apply the classes

clients/web/src/lib/components/atoms/form/Input/Input.svelte
zachstence: Why not support light mode? Its built in to all of the Flowbite components automatically

clients/web/src/lib/components/atoms/form/Input/Input.svelte
zachstence: https://github.com/SprocketBot/sprocket/pull/333/commits/82769efee0558dcc49223eb64f156bf0ecd5adb7

clients/web/src/lib/components/atoms/form/Input/Input.svelte
0c370t: The effort of supporting it both from a design & implementation perspective doesn't make sense (imo) compared to the potential for issues (i.e. a component written in one that doesn't have the responsive styles).

I would guess that the _overwhelming_ majority of our users are likely to use dark mode anyways.

zachstence: Fair, will remove

zachstence: I'd love to support light mode in the future though. Some people (particularly those with disabilities) strongly prefer light mode

zachstence: Fixed https://github.com/SprocketBot/sprocket/pull/333/commits/68996377c29f80fe013b265e8ac4d73998792f1d

0c370t (approved):

PR #344, opened 2022-10-16
feat: show queue banned

bwpikaard:
This is a very temporary solution. Randomly decides not to work and does not auto update. Web rewrite soon:tm:...

image


0c370t:
Probably don't want to merge this if it isn't consistently working.

bwpikaard:
It's a short term fix. /shrug

PR #351, opened 2022-10-22
feat: authorization

bwpikaard:
- Adds an authorization module.
- Scopes organization staff teams and roles to an organization since that's what they are.
- Adds a Guard to check that a member has one of the required actions. Acts as an OR.
- Adds a decorator to require members to have permissions for an action.


gankoji (dismissed):

core/src/authorization/guards/action.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+            context.getClass(),
+        ]);
+
+        if (!requiredActions?.length) return true;
+
+        const ctx = GqlExecutionContext.create(context).getContext();
+        if (!ctx?.req.user) return false;
+
+        const data = JwtAuthPayloadSchema.safeParse(ctx.req.user);
+        if (!data.success || !data.data.currentOrganizationId) return false;
```
zachstence, line 32: We might want a log here if we can't parse the JWT. That shouldn't normally happen, right?

core/src/authorization/guards/action.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        const member = await this.memberRepository.getOrNull({
+            where: {userId: data.data.userId, organizationId: data.data.currentOrganizationId},
+        });
+        if (!member) return false;
+
+        const memberActions = await this.authorizationService.getMemberActions(member.id);
+        if (!requiredActions.some(requiredAction => memberActions.includes(requiredAction))) return false;
```
zachstence, line 40: For this decorator, are `requiredActions` all required, or just one of them is required?

Currently, this `.some` just checks for at least one

core/src/organization/member-restriction/member-restriction.types.ts
```diff
@@ -1,6 +1,8 @@
 import {Field, ObjectType} from "@nestjs/graphql";
 
-import {MemberRestriction} from "$models";
+// For some reason this doesn't exist on $models
+// eslint-disable-next-line no-restricted-imports
+import {MemberRestriction} from "../../database/organization/member_restriction/member_restriction.model";
```
zachstence, line 5: This is weird 🤔 Try to fix?

core/migrations/1666463689317-Automigration.ts
```diff
@@ -0,0 +1,94 @@
...
+        await queryRunner.query(`ALTER TABLE "history"."user_history" DROP COLUMN "type"`);
+        await queryRunner.query(`DROP TYPE "sprocket"."user_type_enum"`);
+        await queryRunner.query(`ALTER TABLE "game_skill_group_profile" DROP CONSTRAINT "FK_238a74250d4d8eb06dbc931cfb4"`);
+        await queryRunner.query(`ALTER TABLE "game_skill_group_profile" ALTER COLUMN "skillGroupId" SET NOT NULL`);
+        await queryRunner.query(`ALTER TABLE "schedule_group_type" DROP CONSTRAINT "FK_fb786f931ab2a1db8b6f8785940"`);
+        await queryRunner.query(`ALTER TABLE "schedule_group_type" ALTER COLUMN "organizationId" SET NOT NULL`);
+        await queryRunner.query(`ALTER TABLE "schedule_fixture" DROP CONSTRAINT "FK_61ef1b6aff93c92ca17d2c64e1d"`);
+        await queryRunner.query(`ALTER TABLE "schedule_fixture" DROP CONSTRAINT "FK_78a0874f503f718434346afedc4"`);
+        await queryRunner.query(`ALTER TABLE "schedule_fixture" ALTER COLUMN "awayFranchiseId" SET NOT NULL`);
+        await queryRunner.query(`ALTER TABLE "schedule_fixture" ALTER COLUMN "homeFranchiseId" SET NOT NULL`);
```
zachstence, line 17: These migrations seem completely unrelated to the code changes... 🤔

zachstence (changes requested):

core/src/authorization/guards/action.guard.ts
bwpikaard: Right. Will do.

core/src/authorization/guards/action.guard.ts
bwpikaard: Just one of them. Acts as an OR. Not sure if that’s what we want but that’s what I assumed.

core/src/organization/member-restriction/member-restriction.types.ts
bwpikaard: I have no clue what the issue is. Importing from $models and logging it just does undefined, but it’s exported everywhere.

bwpikaard: And is only an issue in this file.

core/src/authorization/guards/action.guard.ts
zachstence: That's fine, we should just update the guard interface to be more explicit about that

Maybe something like
```ts
@Action({ oneOf: ["action1", "action2"] })
someFunction(): void {

}
```

zachstence: Then we could also add `allOf`

```ts
@Action({ allOf: ["action1", "action2"] })
someFunction(): void {

}
```

core/src/organization/member-restriction/member-restriction.types.ts
zachstence: If its `undefined`, its likely a circular reference caused by a barrel file somewhere

core/migrations/1666463689317-Automigration.ts
zachstence: Can you create a separate PR for automigrations?

0c370t:
This looks good to me, but this is important enough that I'm going to say it needs tests before we can merge it.

0c370t (approved):

PR #352, opened 2022-10-23
Save salaries to mledb

gankoji:
Salaries were saved to MLEDB only when a rankout was processed for a given player. Add code to save the new salaries even when there's no rankout.


gankoji:
https://app.clickup.com/t/3zjuk4z

bwpikaard (approved):

PR #353, opened 2022-10-23
feat: multi-org part 1

bwpikaard:
- Part 1 of ditching MLEDB
- Allows for multiple organizations to exist without requiring MLE data
- Adds refresh token logic back
- Adds a switchOrgnaization migration to generate a new token for an organization. They do need to be a member.
- User login doesn't require an organization. If they are a member of a single organization, defaults to that organization. Otherwise, is undefined and must be set by switchOrganization.
- Adds Member and Player decorators and guards.
- Scopes scrims to organizations on the matchmaking side instead of core side.


core/src/authentication/authentication.resolver.ts
```diff
@@ -1,16 +1,33 @@
...
+    }
+
+    @Mutation(() => JwtTokenSet)
+    @UseGuards(GraphQLJwtAuthGuard)
+    async switchOrganization(
+        @AuthenticatedUser() user: JwtAuthPayload,
+        @Args("organizationId") organizationId: number,
+    ): Promise<JwtTokenSet> {
+        const organization = await this.organizationRepository.getById(organizationId);
+        return this.authenticationService.login(user.userId, organization.id);
```
0c370t, line 31: Are we checking somewhere that the user is actually in the requested organization?

bwpikaard: Do we want to do that? This will in theory allow us to have an application/form to non members to request membership or something. Will refactor if we want to explicitly say "this person is a member".

0c370t: It looks like it actually does happen in login; but yeah I would say that we do want that. If we want to request membership that can be a user guard, not a member guard.

If any user can switch to any organization I could see that causing issues elsewhere

0c370t (approved):

0c370t (approved):

0c370t (approved):

PR #358, opened 2022-10-25
feat: org switching

bwpikaard:
- Adds the ability to switch between organizations on the front end.

Front end is being rewritten, but mostly for testing before then.


clients/web/src/lib/components/molecules/user-menu/UserMenu.svelte
```diff
@@ -19,6 +19,10 @@
             label: "Sign Out",
             action: async () => goto("/auth/logout"),
         },
+        {
+            label: "Switch Organization",
+            action: async () => goto("/auth/org"),
+        },
```
0c370t, line 25: If the user has multiple organizations, show this; otherwise it should not be shown

clients/web/src/routes/auth/org.svelte
```diff
@@ -0,0 +1,61 @@
...
+
+<script lang="ts">
+    import {goto} from "$app/navigation";
+    import {session} from "$app/stores";
+    import {currentUser} from "$lib/api";
+    import {Card, CenteredCardLayout} from "$lib/components";
+
+    import {switchOrganizationMutation} from "../../lib/api/mutations/SwitchOrganization.mutation";
+    import Spinner from "../../lib/components/atoms/Spinner.svelte";
+    import {type SessionUser, constants, extractJwt} from "../../lib/utils";
```
0c370t, line 21: These should use `$lib` and ensure that we're loading from barrel files properly

clients/web/src/routes/auth/org.svelte
```diff
@@ -0,0 +1,61 @@
...
+            switchOrganization: {access, refresh},
+        } = await switchOrganizationMutation({organizationId: orgId});
+
+        $session.user = extractJwt<SessionUser>(access);
+        $session.token = access;
+
+        cookies.set(constants.auth_cookie_key, access, {expires: 0.25}); // 6 hours
+        cookies.set(constants.refresh_token_cookie_key, refresh, {
+            expires: 7,
+        });
```
0c370t, line 34: Where is this logic done for login elsewhere? We don't want to have 2 different ways of setting these cookies

0c370t (changes requested):

0c370t (approved):

PR #364, opened 2022-10-26
Avatar Atom

mattmccall888:
Avatar atom for sprocket


0c370t:
Task linked: [CU-2q4p54z Avatar Atom](https://app.clickup.com/t/2q4p54z)

clients/web/src/lib/components/atoms/Avatar/Avatar.svelte
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    .shape-square {
+        @apply rounded-none;
+    }
+
+    .shape-circle {
+        @apply rounded-full;
+    }
+    .size-sm{
```
zachstence, line 33: We should have a space before this `{`

Install the Prettier extension in VSCode so that when you save a file, it formats the code nicely :)

image

zachstence (changes requested):

clients/web/src/lib/components/atoms/Avatar/Avatar.svelte
bwpikaard: Use the prettier eslint one, not that one. That doesn’t do eslint fixes. @zachstence

PR #371, opened 2022-10-29
chore: old migration

bwpikaard:


core/migrations/1666463689317-Automigration.ts
```diff
@@ -0,0 +1,94 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Automigration1666463689317 implements MigrationInterface {
+    name = 'Automigration1666463689317'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "user" DROP COLUMN "type"`);
+        await queryRunner.query(`ALTER TABLE "history"."user_history" DROP COLUMN "type"`);
```
0c370t, line 8: Why are we touching history tables here?

bwpikaard: Type is irrelevant and was never used.

bwpikaard: It was something Nigel added in long ago but we ended up using MLE Org Teams.

0c370t: We still shouldn't drop columns from history tables regardless

PR #373, opened 2022-10-29
feat: public module

bwpikaard:
Adds a public module for a public api.


gankoji (approved):

core/src/internal/internal.module.ts
```diff
@@ -20,6 +22,21 @@ import {SubmissionModule} from "./submission";
...
+                    ? {
+                          host: config.redis.host,
+                          servername: config.redis.host,
+                      }
+                    : undefined,
+                keyPrefix: `${config.redis.prefix}:bull`,
+            },
+            prefix: `${config.redis.prefix}:bull`,
+        }),
         AuthenticationModule,
```
zachstence, line 40: Shouldn't we move all of these modules into the `internal/` directory?

zachstence:
Discussion in Discord resulted in deciding not to split

PR #382, opened 2022-11-02
feat: make sure a player can save replays [CU-2jzdzq3]

bwpikaard:


0c370t:
Task linked: [CU-2jzdzq3 Make Sure Someone can Save Replays](https://app.clickup.com/t/2jzdzq3)

core/src/internal/franchise/player/player.service.ts
```diff
@@ -424,4 +424,26 @@ export class PlayerService {
             };
         }
     }
+
+    async playerCanSaveDemos(playerId: number): Promise<boolean> {
+        const player = await this.playerRepository.getById(playerId, {
+            relations: {
+                member: {
+                    platformAccounts: true,
```
0c370t, line 432: Is there a way to filter these into the correct game automatically? Feels like the steps beneath could be done at the database to do less work.

0c370t (approved):

core/src/internal/scrim/scrim.mod.resolver.ts
```diff
@@ -159,6 +160,7 @@ export class ScrimModuleResolver {
         @Args("leaveAfter", {type: () => Int}) leaveAfter: number,
         @Args("group", {nullable: true}) groupKey?: string,
         @Args("createGroup", {nullable: true}) createGroup?: boolean,
+        @Args("canSaveDemos", {type: () => Boolean, nullable: true}) canSaveDemos?: boolean,
```
zachstence, line 163: Why is this an arg to the mutation? Shouldn't we get this info from the implicitly player making the request?

microservices/matchmaking-service/src/scrim/scrim-logic/scrim-logic.service.ts
```diff
@@ -25,24 +27,28 @@ export class ScrimLogicService {
...
+            this.analyticsService
+                .send(AnalyticsEndpoint.Analytics, {
+                    name: "scrimPopped",
+                    strings: [["scrimId", scrim.id]],
+                })
+                .catch(err => {
+                    this.logger.error(err);
+                });
+        } else {
+            await this.scrimService.cancelScrim(scrim.id);
```
zachstence, line 50: We should probably do something other than blindly cancelling. Can we send a notification to players saying why the scrim was cancelled?

zachstence (changes requested):
Instead of cancelling the scrim if nobody can save replays, I think it would be better to prevent somebody from joining if it would fill the scrim with people that can't save replays. I.e., there's 3 console players in a 2s scrim, when a 4th console player tries to join, it doesn't let them, but the other 3 remain in the scrim.

Thoughts?

core/src/internal/scrim/scrim.mod.resolver.ts
bwpikaard: Unsure. Ticket said to add a checkbox for the player that they can save it that will only show up if they can save replays. Use case might be they have accounts that can save demos, but they're playing on one that can't. @0c370t

microservices/matchmaking-service/src/scrim/scrim-logic/scrim-logic.service.ts
bwpikaard: Not through UI unless we add a "cancelled reason" to the scrim cancelled event or if we just do a Discord DM, which we should do but that shouldn't be THE source.

core/src/internal/franchise/player/player.service.ts
bwpikaard: #485f8e1

zachstence (approved):

PR #383, opened 2022-11-08
Breadcrumb atom

0c370t:


clients/web/src/lib/components/atoms/Breadcrumb/Breadcrumb.story.md
```diff
@@ -0,0 +1,9 @@
+# Breadcrumbs
+
+The breadcrumb component is used to show the user where they are in the app.
```
zachstence, line 3: This is cute

clients/web/src/lib/components/atoms/Breadcrumb/addCrumb.ts
```diff
@@ -0,0 +1,4 @@
+import { getContext, setContext } from "svelte";
+import { type ICrumb, type BreadcrumbContext, BREADCRUMB_CONTEXT_KEY } from "./types";
+
+export const addCrumb = (crumb: ICrumb) => setContext<BreadcrumbContext>(BREADCRUMB_CONTEXT_KEY, [...getContext<BreadcrumbContext>(BREADCRUMB_CONTEXT_KEY), crumb])
```
zachstence, line 4: I noticed inside the `Breadcrumb` component, we're using `hasContext` before `getContext`. Is that necessary here? Or is it not necessary in the other place?

clients/web/src/lib/components/atoms/Modal/Modal.story.svelte
```diff
@@ -8,7 +8,7 @@
 </script>
 
 <Hst.Story title="Atoms/Modal" layout={{type: "grid", width: 800}}>
-	<Hst.Variant title="Default" {open}>
+	<Hst.Variant title="Default">
```
zachstence, line 11: Why remove this? Won't this prevent the story controls from working?

zachstence (changes requested):

clients/web/src/lib/components/atoms/Breadcrumb/addCrumb.ts
0c370t: Good catch, it should be used here

clients/web/src/lib/components/atoms/Modal/Modal.story.svelte
0c370t: There aren't any controls for this story; functionality still works.
This was creating type errors in vscode, hence the removal

zachstence (approved):

PR #384, opened 2022-11-08
Drawer Atom

0c370t:


clients/web/src/lib/components/atoms/Drawer/Drawer.svelte
```diff
@@ -0,0 +1,92 @@
...
+            break;
+        case "bottom":
+            flyProps = {y: 100};
+            break;
+    }
+
+    function handleKeyPress(e: KeyboardEvent) {
+        if (!open) return;
+        if (e.key !== "Escape") return;
+        open = false;
```
zachstence, line 33: Stylistic thing, but I think this would be clearer written with just the positive case
We should also use the `close()` function in case some business logic is added there later on

I.e.
```suggestion
        if (open && e.key === "Escape") {
            close()
        }
```

clients/web/src/lib/components/atoms/Drawer/Drawer.svelte
```diff
@@ -0,0 +1,92 @@
...
+            class="fixed z-40 p-4 overflow-y-auto bg-gray-800 drawer {position} text-gray-200"
+            tabindex="-1"
+            transition:fly={flyProps}
+        >
+            <button
+                type="button"
+                class="text-gray-400 bg-transparent rounded-lg text-sm p-1.5 absolute top-2.5 right-2.5 inline-flex items-center hover:bg-gray-600 hover:text-white"
+                on:click={close}
+            >
+                <span class="w-5 h-5"><Icon src={XCircle} /></span>
```
zachstence, line 61: I used the X without a circle to close the Alert here, lets remain consistent
https://github.com/SprocketBot/sprocket/blob/epic/web-rewrite/clients/web/src/lib/components/atoms/Alert/Alert.svelte#L59

zachstence (changes requested):

clients/web/src/lib/components/atoms/Drawer/Drawer.svelte
0c370t: I used X with a circle to close the Modal here :(
https://github.com/SprocketBot/sprocket/blob/epic/web-rewrite/clients/web/src/lib/components/atoms/Modal/Modal.svelte#L34

clients/web/src/lib/components/atoms/Drawer/Drawer.svelte
```diff
@@ -0,0 +1,92 @@
...
+            on:click={close}
+        />
+        <div
+            class="fixed z-40 p-4 overflow-y-auto bg-gray-800 drawer {position} text-gray-200"
+            tabindex="-1"
+            transition:fly={flyProps}
+        >
+            <button
+                type="button"
+                class="text-gray-400 bg-transparent rounded-lg text-sm p-1.5 absolute top-2.5 right-2.5 inline-flex items-center hover:bg-gray-600 hover:text-white"
```
zachstence, line 58: Since this is absolutely positioned, I think it may overlap with the content placed in the `slot`. Can you make a story with some lorem ipsum and make sure the spacing looks good all the way around?

zachstence (changes requested):

zachstence (dismissed):

zachstence (approved):

PR #385, opened 2022-11-09
Toast Atom

0c370t:


clients/web/src/lib/components/abstract/Toast/Toast.svelte
```diff
@@ -0,0 +1,75 @@
...
+            @apply bg-warning-800 text-info-200;
+        }
+        &.danger {
+            @apply bg-danger-800 text-danger-200;
+        }
+    }
+
+    .close {
+        @apply ml-auto rounded-lg focus:ring-2 focus:ring-gray-300 p-1.5 inline-flex h-8 w-8 text-gray-500 hover:text-white bg-gray-800 hover:bg-gray-700;
+    }
```
zachstence, line 74: Move static styles to markdown

clients/web/src/lib/components/abstract/Toast/types.ts
```diff
@@ -0,0 +1,18 @@
+import type {IconSource} from "@steeze-ui/svelte-icon/types";
```
zachstence, line 1: Name this file `toast.types.ts` so its easier to jump to in the project

clients/web/src/lib/components/abstract/Toast/toast.store.ts
```diff
@@ -0,0 +1,29 @@
...
+        id: id,
+        dismissable: !t.ttl,
+        status: ToastStatus.Info,
+        ...t,
+        timestamp: window.performance.now(),
+    };
+
+    _toasts.update(ts => [...ts, newToast]);
+    if (t.ttl) {
+        setTimeout(() => _toasts.update(ts => ts.filter(_t => _t.id !== id)), t.ttl);
```
zachstence, line 22: Call `removeToast` here, defined below

clients/web/src/lib/components/abstract/Toast/toast.store.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+};
+export const removeToast = (id: number) => {
+    _toasts.update(ts => ts.filter(_t => _t.id !== id));
+};
+
+export const toastStore = derived(_toasts, v => v.sort((a, b) => a.id - b.id));
```
zachstence, line 29: Since ID is always incremented, toasts are added at the end of the array, and toasts never change positions, I think this `.sort` is redundant. The toasts should always remain sorted.

zachstence (changes requested):

clients/web/src/lib/components/abstract/Toast/types.ts
0c370t: That would be inconsistent with other type files; you can also use `toast types.ts` or `toast/types.ts` in vs code to jump to the file; shouldn't be an issue.

PR #386, opened 2022-11-09
Select form atom

zachstence:
https://flowbite.com/docs/forms/select/

https://user-images.githubusercontent.com/31896377/200883084-22fba8cb-6c59-474f-8618-a981de8af8ba.mov

### DIFF
--- a/clients/web/src/lib/components/atoms/form/Select/Select.svelte
+++ b/clients/web/src/lib/components/atoms/form/Select/Select.svelte
@@ -41,12 +41,13 @@
 </script>
 
 <div class="size-{size} state-{state}" class:disabled>
-    <label for={labelId}>{label}</label>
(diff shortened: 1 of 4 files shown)


clients/web/src/lib/components/atoms/form/Select/Select.svelte
```diff
@@ -0,0 +1,159 @@
...
+        @apply relative;
+    }
+
+    .icon {
+        @apply absolute h-full top-0 flex items-center pointer-events-none;
+    }
+
+    .disabled {
+        @apply text-gray-400;
+    }
```
0c370t, line 89: Should be moved to `class` attributes

zachstence: [998169d](https://github.com/SprocketBot/sprocket/pull/386/commits/998169dd32cdee569156d2748adf2862c2e99755)

0c370t (approved):

PR #390, opened 2022-11-11
feat: scrim locked reason [CU-3yw11q7]

MattRusselli:
adds a reason for why scrims are being locked

### DIFF
--- a/microservices/matchmaking-service/src/scrim/scrim.controller.ts
+++ b/microservices/matchmaking-service/src/scrim/scrim.controller.ts
@@ -92,7 +92,10 @@ export class ScrimController {
     @MessagePattern(MatchmakingEndpoint.SetScrimLocked)
     async setScrimLocked(@Payload() payload: unknown): Promise<boolean> {
         const data = MatchmakingSchemas.SetScrimLocked.input.parse(payload);
-        await this.scrimService.setScrimLocked(data.scrimId, data.locked);
+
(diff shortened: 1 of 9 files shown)


0c370t:
Task linked: [CU-3yw11q7 Scrim Locked Reason](https://app.clickup.com/t/3yw11q7)

zachstence (commented):
Beautiful PR, nice job!

Are there any opportunities where this locked reason is going to be used? If so, can we implement that in this PR?

microservices/matchmaking-service/src/scrim/scrim.controller.ts
```diff
@@ -92,7 +92,10 @@ export class ScrimController {
     @MessagePattern(MatchmakingEndpoint.SetScrimLocked)
     async setScrimLocked(@Payload() payload: unknown): Promise<boolean> {
         const data = MatchmakingSchemas.SetScrimLocked.input.parse(payload);
-        await this.scrimService.setScrimLocked(data.scrimId, data.locked);
+
+        if (data.locked) await this.scrimService.setScrimLocked(data.scrimId, data.locked, data.reason);
+        else await this.scrimService.setScrimLocked(data.scrimId, data.locked);
+
```
0c370t, line 98: This change may not be needed; if data.reason is undefined we can safely pass it.

bwpikaard: It errored that it didn’t exists because of the union and if being false, TS didn’t like it.

PR #394, opened 2022-11-13
Card Atom

gankoji:


clients/web/src/lib/components/atoms/Card/Card.story.svelte
```diff
@@ -0,0 +1,29 @@
+<script lang="ts">
+    import type {Hst as _Hst} from "@histoire/plugin-svelte";
+    import Card from "./Card.svelte";
+
+    type sizes = "small" | "default" | "large" | "extraLarge";
+
+    const sizeOptions = ["small", "default", "large", "extraLarge"];
+
```
0c370t, line 8: Is this dead code?

clients/web/src/lib/components/atoms/Card/Card.svelte
```diff
@@ -0,0 +1,45 @@
...
+                viewBox="0 0 20 20"
+                xmlns="http://www.w3.org/2000/svg"
+                ><path
+                    fill-rule="evenodd"
+                    d="M10.293 3.293a1 1 0 011.414 0l6 6a1 1 0 010 1.414l-6 6a1 1 0 01-1.414-1.414L14.586 11H3a1 1 0 110-2h11.586l-4.293-4.293a1 1 0 010-1.414z"
+                    clip-rule="evenodd"
+                /></svg
+            >
+        </a>
+    {/if}
```
0c370t, line 44: We can likely rip some (or all) of this out and just replace it with a `<slot/>` (or 2, one for the header if we feel it's needed); to keep this a little more agnostic

0c370t (changes requested):

clients/web/src/lib/components/atoms/Card/Card.story.svelte
gankoji: Yes.

clients/web/src/lib/components/atoms/Card/Card.svelte
gankoji: Done.

0c370t (approved):

PR #406, opened 2022-12-07
Make local environment setup tenable

ItsMeBrianD:


common/src/util/config.ts
```diff
@@ -1,203 +1,301 @@
...
             get secret(): string {
-                return readFileSync("./secret/googleSecret.txt").toString();
+                if (_config.has("auth.google.secret")) {
+                    return _config.get("auth.google.secret");
+                }
+                return readSecretFile("./secret/googleSecret.txt");
             },
             get callbackUrl(): string {
-                return _config.get<string>("auth.google.callbackUrl");
+                return _config.get("auth.google.callbackUrl");
```
gankoji, line 41: This shows that we are, at minimum, forcing the user to have a config json file. Is that the design? Makes sense, one json for non-secret stuff and one env for the secrets.

gankoji (approved):

ItsMeBrianD:
This effort is moving to v2

PR #409, opened 2022-12-09
Tooltip Atom / Action

0c370t:


clients/web/src/lib/components/actions/Tooltip/Tooltip.story.svelte
```diff
@@ -0,0 +1,35 @@
...
+            <span class="ml-8">With Arrow</span>
+            <input bind:checked={params.withArrow} type="checkbox" />
+        </div>
+    </svelte:fragment>
+    <Hst.Variant title="Default">
+        <div class="w-screen h-screen flex justify-center items-center">
+            {#key params}
+                <button class="border border-primary rounded-lg px-2 py-1 text-gray-200" use:tooltip={params}
+                    >{params.position}</button
+                >
```
zachstence, line 31: Funky formatting here

clients/web/src/lib/components/actions/Tooltip/Tooltip.story.svelte
```diff
@@ -0,0 +1,35 @@
...
+            </select>
+        </div>
+        <div>
+            <span class="ml-8">With Arrow</span>
+            <input bind:checked={params.withArrow} type="checkbox" />
+        </div>
+    </svelte:fragment>
+    <Hst.Variant title="Default">
+        <div class="w-screen h-screen flex justify-center items-center">
+            {#key params}
```
zachstence, line 28: What does this `{#key}` do? Never seen this before

clients/web/src/lib/components/actions/Tooltip/useTooltip.action.ts
```diff
@@ -0,0 +1,63 @@
+import type {TooltipParams} from "./types";
+import "./tooltip.postcss";
+
+export const tooltip = (node: HTMLElement, {content, position = "top", withArrow}: TooltipParams) => {
+    const tooltip = document.createElement("div");
+    tooltip.classList.add("tooltip");
+    console.log({position, withArrow});
```
zachstence, line 7: Remove log

clients/web/src/lib/components/actions/Tooltip/useTooltip.action.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+    node.addEventListener("mouseenter", renderTooltip);
+    node.addEventListener("mouseleave", removeTooltip);
+
+    return {
+        destroy() {
+            node.removeEventListener("mouseenter", renderTooltip);
+            node.removeEventListener("mouseleave", removeTooltip);
+        },
+    };
```
zachstence, line 62: What is this syntax? Is this returning an object with a function `destroy`? No function keyword needed?

clients/web/src/lib/components/actions/Tooltip/tooltip.postcss
```diff
@@ -0,0 +1,25 @@
+.tooltip {
```
zachstence, line 1: Are these styles global? If so, we may need a more unique class name than `tooltip`. I could see that being duplicated somewhere else in the app

zachstence (changes requested):
`src/lib/components/atoms/Tooltip/index.ts` is an empty file

clients/web/src/lib/components/actions/Tooltip/Tooltip.story.svelte
0c370t: This is what prettier does in some cases 😞 
I'll try to tweak it

clients/web/src/lib/components/actions/Tooltip/Tooltip.story.svelte
0c370t: Forces a re-render when the prop changes; in this case it forces a rerender when parameters change, I got funky behavior without it

clients/web/src/lib/components/actions/Tooltip/useTooltip.action.ts
0c370t:
```suggestion
```

clients/web/src/lib/components/actions/Tooltip/useTooltip.action.ts
0c370t: Function keyword isn't technically ever needed.

This does return that object; see https://svelte.dev/docs#template-syntax-element-directives-use-action

clients/web/src/lib/components/actions/Tooltip/tooltip.postcss
0c370t: Not sure what other context `tooltip` would be showing up in, but I believe this is global

zachstence (approved):

PR #410, opened 2022-12-09
Tabs Atom

0c370t:

### DIFF
--- a/clients/web/src/lib/components/atoms/Tabs/Tabs.svelte
+++ b/clients/web/src/lib/components/atoms/Tabs/Tabs.svelte
@@ -1,0 +1,46 @@
+<script lang="ts">
+    import type {TabItems} from "./types";
+    import {Icon} from "@steeze-ui/svelte-icon";
+
+    type T = $$Generic<TabItems>;
+    interface $$Props {
+        tabs: TabItems;
+        activeKey: keyof TabItems;
+    }
+
+    export let tabs: TabItems;
+    export let activeKey: keyof TabItems;
+</script>
+
+<div class="text-sm font-medium text-center border-b text-gray-400 border-gray-700">
+    <ul class="flex flex-wrap -mb-px">
+        {#each Object.keys(tabs) as key}
+            <li class="mr-2">
+                <button
+                    type="button"
+                    class="inline-block p-2 rounded-t-lg border-2 border-transparent hover:text-gray-600 hover:border-b-gray-300 hover:text-gray-300 flex items-center gap-2"
+                    class:active={key === activeKey}
+                    on:click={() => (activeKey = key)}
+                >
+                    {#if tabs[key].icon}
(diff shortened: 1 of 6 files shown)


clients/web/src/lib/components/atoms/Tabs/types.ts
```diff
@@ -0,0 +1,18 @@
+import type {SvelteComponent, ComponentProps} from "svelte";
+import type {IconSource} from "@steeze-ui/svelte-icon/types";
+
+// export type TabItem = {
```
zachstence, line 4: Remove comment

clients/web/src/lib/components/atoms/Tabs/types.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type TabItem<T extends SvelteComponent = SvelteComponent> = {
+    component: T;
+    props: ComponentProps<T>;
+    icon?: IconSource;
+};
+
+export type TabItems = {
+    [key: string]: TabItem;
+};
```
zachstence, line 18: Can be simplified

```suggestion
export type TabItems = Record<string, TabItem>;
```

clients/web/src/lib/components/atoms/Tabs/Tabs.svelte
```diff
@@ -0,0 +1,46 @@
...
+    interface $$Props {
+        tabs: TabItems;
+        activeKey: keyof TabItems;
+    }
+
+    export let tabs: TabItems;
+    export let activeKey: keyof TabItems;
+</script>
+
+<div class="text-sm font-medium text-center border-b text-gray-400 border-gray-700">
```
zachstence, line 15: There are some accessibility considerations we need to implement

See
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tabpanel_role

For implementation, the official `react-tabs` library would be good to follow (there are several `aria` properties we should implement)

zachstence (changes requested):

clients/web/src/lib/components/atoms/Tabs/types.ts
```diff
@@ -0,0 +1,18 @@
+import type {SvelteComponent, ComponentProps} from "svelte";
+import type {IconSource} from "@steeze-ui/svelte-icon/types";
+
+// export type TabItem = {
+//   component: SvelteComponent,
+//   key: string,
+//   props: ComponentProps<SvelteComponent>
+// }
+
```
0c370t, line 9:
```suggestion
```

zachstence (approved):

PR #418, opened 2022-12-13
Refactor: Move Platform Accounts from Member to User [CU-422ucek]

bwpikaard:
TODO: Make changes to finalization and player service.
TODO: import from MLEDB player accounts instead because sprocket has a lot of duplicates from mledb->sprocket sync and names and IDs both existing. Or delete and resync.


0c370t:
Task linked: [CU-422ucek Migrate Member Platform Accs to User](https://app.clickup.com/t/422ucek)

PR #421, opened 2022-12-17
Atom: Buttons

gankoji:


clients/web/src/lib/components/atoms/Button/AltButton.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+    export let label:string;
+</script>
+
+<button type="button" class="text-white bg-secondary-700 hover:bg-primary-800 
+focus:ring-4 focus:ring-primary-300 font-medium rounded-lg text-sm 
+px-5 py-2.5 mr-2 mb-2 focus:outline-none">Alternative</button>
```
zachstence, line 7: Instead of having a `AltButton` component, lets combine it with `Button` and have a `variant` prop that can either be `"primary"` or `"secondary"`, and defaults to `"primary"`

clients/web/src/lib/components/atoms/Button/Button.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+    export let label:string;
```
zachstence, line 2: Lets accept the label via the default slot rather than a prop

See: https://svelte.dev/repl/a295667e0f384c6da13d58f92a24d934?version=3.55.0

clients/web/src/lib/components/atoms/Button/Button.svelte
```diff
@@ -0,0 +1,7 @@
...
+</script>
+
+<button type="button" class="text-white bg-primary-700 hover:bg-secondary-800
```
zachstence, line 5: Needs to handle `on:click`

See https://svelte.dev/repl/a295667e0f384c6da13d58f92a24d934?version=3.55.0

zachstence (changes requested):

clients/web/src/lib/components/atoms/Button/AltButton.svelte
gankoji: Done.

clients/web/src/lib/components/atoms/Button/Button.svelte
gankoji: Done.

clients/web/src/lib/components/atoms/Button/Button.svelte
gankoji: Done.

zachstence (approved):

PR #426, opened 2023-01-06
Create PostgreSQL trigger to synchronize Sprocket -> MLEDB salaries

mwbrown:
This query, when run, will create a trigger on the `sprocket.player` table every time a player's salary is updated. It will map the Sprocket player ID to the respective MLE player ID using `mledb_bridge.player_to_player`, then update the salary for the corresponding player in the `mledb.player` table.


0c370t:
We may need to investigate SQL formatting solutions, but this is looking good; one request, if we can add this as a typeorm migration so it is automatically added to the database and table; that would be great.

Example here:
https://github.com/SprocketBot/sprocket/blob/dev/core/migrations/1658698707748-AuditingFunctions.ts

0c370t (approved):

PR #435, opened 2023-01-10
fix: intake

bwpikaard:


core/src/franchise/player/player.resolver.ts
```diff
@@ -158,8 +159,6 @@ export class PlayerResolver {
         });
         const orgProfile = await this.organizationService.getOrganizationProfileForOrganization(player.member.organization.id);
 
-        if (player.skillGroup.id === skillGroupId) return "ERROR: This player is already in this skill group";
```
0c370t, line 161: Should this be a warn instead of an error/throwing it away?

core/src/franchise/player/player.service.ts
```diff
@@ -301,6 +301,7 @@ export class PlayerService {
     ): Promise<MLE_Player> {
         const updatedPlayer = this.mle_playerRepository.merge(player, {
             updatedBy: "Sprocket FA Intake",
+            updatedAt: new Date(),
```
0c370t, line 304: I feel like this should be automatic, might be worth looking at the model and making sure the property is decorated correctly

0c370t (approved):

PR #451, opened 2023-01-29
feat: default token

bwpikaard:
Allows you to set a token in your configuration file to authenticate for all requests without needing to set a token. Useful for testing subscriptions because you cannot provide authentication tokens in GQL Playground or Altair. Defaults to the token in the header before trying to use the config token so you can have "multiple users".


gankoji (dismissed):

bwpikaard:
Do not merge until parent branch is merged to dev.

0c370t (approved):

PR #464, opened 2023-02-23
fix: authentication issues

bwpikaard:


core/src/authentication/authentication.controller.ts
```diff
@@ -36,13 +36,21 @@ export class AuthenticationController {
     @UseGuards(GoogleAuthGuard)
     async googleLogin(@Request() req: Req, @Response() res: Res, @Query("state") state?: string): Promise<void> {
         const googleProfile = GoogleProfileSchema.safeParse(req.user);
-        if (!googleProfile.success) throw new HttpException("Could not validate user", 400);
+        if (!googleProfile.success) {
+            res.redirect(`${config.auth.frontend_callback}?status=error&message=${"Failed to parse provider profile"}`);
```
0c370t, line 40: This error message needs to be more user friendly; we can log information to give developers an idea of what went wrong (can generate some id to associate the user's message with the log message), but this doesn't tell an end user anything

core/src/authentication/authentication.controller.ts
```diff
@@ -36,13 +36,21 @@ export class AuthenticationController {
...
+        }
 
         if (state) {
             try {
                 const payload = this.jwtService.verify(state);
                 const data = JwtAuthPayloadSchema.safeParse(payload);
-                if (!data.success) throw new HttpException("Could not validate user", 400);
+                if (!data.success) {
+                    res.redirect(
+                        `${config.auth.frontend_callback}?status=error&message=${"Failed to parse jwt payload"}`,
```
0c370t, line 50: ditto

core/src/authentication/authentication.controller.ts
```diff
@@ -57,22 +65,27 @@ export class AuthenticationController {
                         userId: user.id,
                     });
             } catch (e) {
+                res.redirect(`${config.auth.frontend_callback}?status=error&message=${(e as Error).message}`);
```
0c370t, line 68: We might want to be careful presenting error messages in this way

core/src/authentication/authentication.controller.ts
```diff
@@ -57,22 +65,27 @@ export class AuthenticationController {
...
+                    });
                 const tokens = await this.authenticationService.login(user.id);
 
-                res.redirect(`${config.auth.frontend_callback}?token=${tokens.access},${tokens.refresh}`);
+                res.redirect(
+                    `${config.auth.frontend_callback}?token=${tokens.access},${tokens.refresh}&status=success`,
+                );
                 return;
             } catch (e) {
+                res.redirect(`${config.auth.frontend_callback}?status=error&message=${(e as Error).message}`);
```
0c370t, line 86: ditto

PR #472, opened 2023-03-11
Personal/gankoji/onboarding part one of many

gankoji:
This bash script simply creates the `core/secret` directory and populates it with the necessary files and dummy data to get Core to not fall over on startup. You will still have to add in the credentials manually.


mwbrown (approved):

ItsMeBrianD:
This should be good to go once merge conflicts are resolved

ItsMeBrianD:
How does this relate to #406?

PR #478, opened 2023-03-19
feat: login as user

bwpikaard:


core/src/identity/identity.module.ts
```diff
@@ -10,7 +13,15 @@ import {
 import {UserAuthenticationAccountResolver} from "./user-authentication-account";
 
 @Module({
-    imports: [DatabaseModule, UtilModule],
+    imports: [
+        DatabaseModule,
+        UtilModule,
+        JwtModule.register({
+            secret: readFileSync("./secret/jwtSecret.txt").toString()
```
ItsMeBrianD, line 20: This should be moved to config in common

core/src/identity/user/user.resolver.ts
```diff
@@ -74,4 +80,18 @@ export class UserResolver {
...
+
+    @UseGuards(GqlJwtGuard, MLEOrganizationTeamGuard(MLE_OrganizationTeam.MLEDB_ADMIN))
+    @Mutation(() => String)
+    async loginAsUser(@Args("userId", {type: () => Int}) userId: number): Promise<string> {
+        const user = await this.userService.getUserById(userId, {relations: {profile: true} });
+        const payload: AuthPayload = {
+            sub: `${user.id}`,
+            username: user.profile.displayName,
+            userId: user.id,
+            currentOrganizationId: config.defaultOrganizationId,
```
ItsMeBrianD, line 92: Use optional arg with default to this

ItsMeBrianD (changes requested):

gankoji (dismissed):

ItsMeBrianD (approved):

PR #486, opened 2023-04-21
fix: markReplaysNCP didn't set winning team name

gankoji:


mwbrown (approved):
Let's Gamble, Try Merging

PR #494, opened 2023-07-23
Personal/gankoji/bulk elo updates

gankoji:


core/src/franchise/player/player.resolver.ts
```diff
@@ -238,6 +238,67 @@ export class PlayerResolver {
...
+        for (const player of parsed) {
+            try {
+                await this.changePlayerElo(player.playerId, player.salary, player.newElo);
+            } catch {
+                numFailed++;
+                continue;
+            }
+        }
+        
+        return (numFailed === 0) ? `Success, all elos changed.` : `${numFailed} elos were unable to be changed.`;
```
bwpikaard, line 299: Return a list of who failed, not just a number so we can look at who/why and not need to guess.

bwpikaard (changes requested):

core/src/franchise/player/player.resolver.ts
gankoji: Done.

ItsMeBrianD (approved):

PR #507, opened 2023-10-14
Fix: Some franchise staff unable to submit or ratify matches

gankoji:
Bug was discovered by MLE user Jdogg, who is a blizzard captain but couldn't submit/ratify. Logs showed:

image

Jdogg is one of a handful of captains who have this issue. Discovered that all players with this issue (correct perms, but still unable) had a mismatch between their Sprocket UserId and MemberId (not designed to match, but by happenstance often do for 80%+ of the current userbase). Explicitly getting UserId from the Member table for a given player fixes this.

### DIFF
--- a/core/src/franchise/franchise/franchise.service.ts
+++ b/core/src/franchise/franchise/franchise.service.ts
@@ -41,7 +41,7 @@ export class FranchiseService {
     memberId: number
   ): Promise<CoreOutput<CoreEndpoint.GetPlayerFranchises>> {
     const member = await this.sprocketMemberService.getMemberById(memberId);
-    const userId = member.userId;
+    const { userId } = member;
     const mlePlayer = await this.mledbPlayerService.getMlePlayerBySprocketUser(
       userId
     );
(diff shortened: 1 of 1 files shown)


ItsMeBrianD:
Looks like this branch is polluted, is that intentional?

gankoji:
> Looks like this branch is polluted, is that intentional?

fixed XD

core/src/franchise/franchise/franchise.service.ts
```diff
@@ -1,65 +1,80 @@
...
+  async getFranchise(query: FindOneOptions<Franchise>): Promise<Franchise> {
+    return this.franchiseRepository.findOneOrFail(query);
+  }
 
-        const playerId = mlePlayer.id;
+  async getPlayerFranchises(
+    memberId: number
+  ): Promise<CoreOutput<CoreEndpoint.GetPlayerFranchises>> {
+    const member = await this.sprocketMemberService.getMemberById(memberId);
+    const userId = member.userId;
```
ItsMeBrianD, line 44: Not blocking but could do

```suggestion
    const { userId } = member;
```

ItsMeBrianD (approved):

ItsMeBrianD (approved):

PR #511, opened 2023-10-24
fix: Not FPs get called FP

gankoji:
in all places where we try to get the franchises attached to a sprocket User, do so with their *memberId* in the organization.


ItsMeBrianD (approved):
oof

PR #544, opened 2024-07-14
Add multi-game support for user intake

gankoji:
- Refactor player service (particularly the create MLE player function for default args)
- Add new mutation in user resolver, which:
  - Registers a user and UAA
  - Creates a member within MLE
  - Loops over the input list of tuples of `(skill_group, salary)` to create a player for each corresponding game
  - Creates a single MLE player with the last salary and player ID from that loop

### DIFF
--- a/core/src/identity/user/user.resolver.ts
+++ b/core/src/identity/user/user.resolver.ts
@@ -5,12 +5,15 @@ import {
 import {JwtService} from "@nestjs/jwt";
 import {config} from "@sprocketbot/common";
 
-import type {UserAuthenticationAccount, UserProfile} from "../../database";
+import type {
+    UserAuthenticationAccount, UserProfile,
+} from "../../database";
 import {
     Member, User, UserAuthenticationAccountType,
 } from "../../database";
 import {MLE_OrganizationTeam} from "../../database/mledb";
 import {PlayerService} from "../../franchise";
+import {IntakePlayerAccount} from "../../franchise/player/player.resolver";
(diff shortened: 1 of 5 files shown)


XsNIPER50 (approved):
Looks good.

core/src/identity/user/user.resolver.ts
```diff
@@ -19,13 +25,17 @@ import {GqlJwtGuard} from "../auth/gql-auth-guard";
 import {IdentityService} from "../identity.service";
 import {UserService} from "./user.service";
 
+const MLE_ORGANIZATION_ID = 2;
```
ItsMeBrianD, line 28: We need to avoid magic numbers like this if at all possible

core/src/identity/user/user.resolver.ts
```diff
@@ -103,4 +113,47 @@ export class UserResolver {
...
+    ): Promise<User> {
+        // This looks a little backwards, but the identity service actually
+        // creates the user object *and* the associated auth account for login
+        // at the same time.
+        const user = await this.identityService.registerUser(UserAuthenticationAccountType.DISCORD, d_id);
+        
+        // Once we have the user, we can create the member as part of MLE
+        const member = await this.memberService.createMember(
+            {name: name},
+            MLE_ORGANIZATION_ID,
```
ItsMeBrianD, line 133: We could look this up just prior to usage

ItsMeBrianD (approved):
I'm not super thrilled with the hard coding here; but would like to get this across the line

PR #554, opened 2024-12-17
Improve Matchmaking for 3s RR Scrims

gankoji:
Initial thought was to go with Ruskey's algorithm for generating unique set partitions of a given size: this turned out to be unnecessarily complicated, so I fell back to a really basic power set expansion and filtered for the ones we wanted. It's horrible time complexity (literally O(2^n)), but n is 6 and we only do it on initialization. 

TL;DR - Fix the 'two people get stuck with each other each 3s scrim' problem.

### DIFF
--- a/microservices/matchmaking-service/src/scrim/game-order/game-order.service.ts
+++ b/microservices/matchmaking-service/src/scrim/game-order/game-order.service.ts
@@ -3,18 +3,72 @@ import type {
     Scrim, ScrimGame, ScrimPlayer,
     ScrimTeam,
 } from "@sprocketbot/common";
-import {ScrimMode} from "@sprocketbot/common";
-import shuffle from "lodash.shuffle";
+import {ScrimMode, ScrimStatus} from "@sprocketbot/common";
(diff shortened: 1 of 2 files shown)


microservices/matchmaking-service/src/scrim/game-order/game-order.service.ts
```diff
@@ -3,18 +3,72 @@ import type {
...
     ScrimTeam,
 } from "@sprocketbot/common";
-import {ScrimMode} from "@sprocketbot/common";
-import shuffle from "lodash.shuffle";
+import {ScrimMode, ScrimStatus} from "@sprocketbot/common";
+import { add, sub } from "date-fns";
+import { now } from "lodash";
 
 import {ScrimGroupService} from "../scrim-group/scrim-group.service";
-import {createCombinations} from "./combinations";
```
ItsMeBrianD, line 10: Is this dead code now?

ItsMeBrianD (dismissed):
Seems good

ItsMeBrianD (approved):

PR #558, opened 2024-12-24
Casbin RBAC!

gankoji:
This time, with decorators!

Default perms are a little goofy. This really needs a web interface for managing perms. That will come next.


core/src/app.module.ts
```diff
@@ -42,8 +100,17 @@ import { HealthModule } from './health/health.module';
...
+      userFromContext: (ctx) => {
+        const req = ctx.getArgs()[2]['request'];
+        const user = getUserFromRequest(req);
+
+        if (user) {
+          Logger.log(`Found user from request: ${JSON.stringify(user)}.`);
+          return user.id;
+        } else {
+          return 'anonymous';
+        }
```
ItsMeBrianD, line 112: Doesn't this need to return a full user? a la https://github.com/SprocketBot/sprocket/blob/v2/lib/types/auth/user.ts#L11-L16

core/src/auth/auth.controller.ts
```diff
@@ -39,4 +43,32 @@ export class AuthController {
...
+    // Give superuser access to everything
+    (Object.keys(Resource) as Array<keyof typeof Resource>).map(
+      async (resource) => {
+        (Object.keys(ResourceAction) as Array<keyof typeof ResourceAction>).map(
+          async (action) => {
+            this.addPermTryCatch('superuser', resource, action + ':any');
+          },
+        );
+      },
+    );
```
ItsMeBrianD, line 58: We should add this to the seed script rather than having an endpoint for it; see here:
https://github.com/SprocketBot/sprocket/blob/v2/core/src/db/seed.module.ts#L20-L56

core/src/auth/auth.controller.ts
```diff
@@ -39,4 +43,32 @@ export class AuthController {
...
+
+    // Make Nigel a superuser
+    await this.authZService.addRoleForUser(
+      'ce5a3b01-f81c-4391-807c-4be84336ab19',
```
ItsMeBrianD, line 62: I think we have a configuration for this (also hard-coded guid needs to be removed)

https://github.com/SprocketBot/sprocket/blob/v2/config.yaml#L35-L37

core/src/auth/auth.controller.ts
```diff
@@ -39,4 +43,32 @@ export class AuthController {
...
+    );
+  }
+
+  async addPermTryCatch(user: string, resource: string, action: string) {
+    try {
+      await this.authZService.addPermissionForUser(user, resource, action);
+    } catch {
+      this.logger.warn('Permission already exists');
+    }
+  }
```
ItsMeBrianD, line 73: `ensurePerm`?

core/src/db/authorize/constants.ts
```diff
@@ -0,0 +1,24 @@
...
+  TeamStat = 'team_stat',
+  User = 'user',
+  UserAuthAccount = 'user_auth_account',
+}
+
+export enum ResourceAction {
+  Create = 'create',
+  Read = 'read',
+  Update = 'update',
+  Delete = 'delete',
```
ItsMeBrianD, line 23: There's a section in `lib` that has items like this so that they can be shared between clients and the backend, might be good to put stuff there to ease interacting with it on FE

core/src/gql/game/game.resolver.ts
```diff
@@ -3,14 +3,21 @@ import { SkillGroupObject } from '../skill_group/skill_group.object';
...
 export class GameResolver {
   constructor(private readonly gameRepo: GameRepository) {}
 
   @Query(() => [GameObject])
-  @UseGuards(AuthorizeGuard())
+  @UseGuards(AuthZGuard)
+  @UsePermissions({
+    resource: Resource.Game,
+    action: ResourceAction.Read,
+    possession: AuthPossession.ANY,
```
ItsMeBrianD, line 19: More of a question, but can we extend this possession concept to properly account for franchise / team / league scopings?

ItsMeBrianD (changes requested):
Looks pretty good, a few small refactor / code location things

core/src/gql/game/game.resolver.ts
gankoji: I'm fairly certain this possession concept will work for that purpose as is, but we can always add another field to the policy if we find we need another dimension.

core/src/app.module.ts
gankoji: It looks like both ways are supported, as long as the returned object has a 'username' field. https://github.com/node-casbin/nest-authz/blob/master/README.md

core/src/auth/auth.controller.ts
gankoji: Fixed, now loops over those names in the config.

core/src/auth/auth.controller.ts
gankoji: Fixed, seeding is done in seed module instead of wonky route.

core/src/auth/auth.controller.ts
gankoji: Fixed.

core/src/db/authorize/constants.ts
gankoji: Fixed.

bun.lockb
ItsMeBrianD, line 1: Sussed out that this seems to be down 96%

ItsMeBrianD (approved):
Changes look good, should just be a peek at the lock file and then this is ready to merge :rocket:

PR #562, opened 2025-02-15
feat: Validate matches follow substitution rules.

gankoji:
Due to the flexibility allowed by MLE's substitution rules, we can't validate that *only* the correct number of players played in a given 2s or 3s match, so until now we haven't been able to prevent folks from submitting 3s replays as 2s matches and vice versa. 

However, what we can do is require the substitution rules are followed *exactly*, which gets us much closer:

- Only one substitution per match per team
- Which means at most 3(4) unique players per team
- AND, at most 1 replay in which a given team had more than 2(3) players

We can also check that the opposites applied:
- At *minimum* 2(3) unique players per team

This takes the gameModeId given to replay-validation from the core endpoint and uses it to switch the appropriate thresholds. Then, we call the new method `validateTeams` to apply the checks listed above. Should significantly reduce the number of NCPs needed due to poor reading comprehension.

### DIFF
--- a/microservices/submission-service/src/replay-validation/replay-validation.service.ts
+++ b/microservices/submission-service/src/replay-validation/replay-validation.service.ts
@@ -5,7 +5,8 @@ import type {
     LFSReplaySubmission,
     MatchReplaySubmission,
     ReplaySubmission,
-    ScrimPlayer,    ScrimReplaySubmission,
+    ScrimPlayer,
+    ScrimReplaySubmission,
     UpdateLFSScrimPlayersRequest,
 } from "@sprocketbot/common";
 import {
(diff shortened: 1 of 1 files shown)


matthieu-trudel:
i just want to comment on the "AND, at most 1 replay in which a given team had more than 2(3) players"

LO rulebook, rule 6.11 states "Mid-game substitutions are never allowed."

follow up question, should we be prompting the user about the error, giving them the opportunity to fix it (which is what is currently in place) but also allow them to accept the replays as "yes this happened, yes we'll take the NCP"?

gankoji:
From MLE's perspective, we don't want NCPs to be handled by anyone but LO, even when it's an obvious case. There is more to the NCP than just "ok we know it's wrong/we didn't follow the rules," so I'm content to leave it without intervention being possible here. I think that NCP logic in general is very-much league specific, and thus should live in consumer services and not the Sprocket platform. (I know that this directly refutes what I've just done with the replay parse logic, but what I'm saying is that was pretty much the last exception to this rule that I'm willing to make, because it's 1.0 and we're throwing it out anyway)

PR #574, opened 2025-05-27
Gql classes update

matthieu-trudel:
just some small update to some of the classes
and added myself as an initialAdmin


core/src/db/club/club.entity.ts
```diff
@@ -1,5 +1,8 @@
 import { BaseEntity } from '../base.entity';
-import { Entity } from 'typeorm';
+import { Column, Entity } from 'typeorm';
 
 @Entity('club', { schema: 'sprocket' })
-export class ClubEntity extends BaseEntity {}
+export class ClubEntity extends BaseEntity {
+	@Column()
+	franchiseId: string;
```
gankoji, line 7: This should either just be type Franchise, or have both Franchise and franchiseId. The important bit is the relation to the Franchise entity via TypeORM, in this case via OneToMany

core/src/db/team/team.entity.ts
```diff
@@ -2,4 +2,6 @@ import { BaseEntity } from '../base.entity';
 import { Entity } from 'typeorm';
 
 @Entity('team', { schema: 'sprocket' })
-export class TeamEntity extends BaseEntity {}
+export class TeamEntity extends BaseEntity {
+	clubId: string;
```
gankoji, line 6: Same comment here. If you need an example, see player entity.

core/src/gql/franchise/franchise.object.ts
```diff
@@ -2,15 +2,15 @@ import { Field, ObjectType } from '@nestjs/graphql';
 
 @ObjectType('Franchise')
 export class FranchiseObject {
-  @Field()
-  id: string;
+	@Field()
```
gankoji, line 5: Not necessary to address in this PR, but leaving it as a reminder: we should have a standard formatter for this project and it should be run as a githook. I should not be having to review whitespace, as much as possible (all of us shouldn't, not just me!) 
image

core/src/gql/franchise/franchise.resolver.ts
```diff
@@ -1,10 +1,16 @@
...
-  @ResolveField()
-  async getFranchise(): Promise<FranchiseObject> {
-    throw new Error('Not yet implemented');
-  }
+	constructor(private readonly franchiseRepo: FranchiseRepository) {}
+	@Query()
+	async franchise(@Args('id') id: string, @Args('name') name?: string): Promise<FranchiseObject> {
+		const franchise = await this.franchiseRepo.findOneBy({
+			id,
+			name
```
gankoji, line 12: Does `findOneBy` handle null arguments?

core/src/gql/franchise/franchise.object.ts
```diff
@@ -2,15 +2,15 @@ import { Field, ObjectType } from '@nestjs/graphql';
...
+	id: string;
 
-  @Field()
-  createdAt: Date;
+	@Field()
+	createdAt: Date;
```
gankoji, line 9: Another comment that is very addressable in a separate PR. Is it not possible for us to have a BaseObject like we do with BaseEntity?

gankoji (changes requested):

matthieu-trudel:
closing this PR; breaking it down in the following PRs

PR #575, opened 2025-06-10
BaseObject implementation

matthieu-trudel:
just to get it out of the way. the rest of the implementation is in the gqlupdate branch

only thing that should be here is the bare minimum (removal of id, createdAt and updatedAt fields and transferring those into the BaseObject, and then extending that). had to create the base_seat_assignment entity as it was missing for the base_seat_assignment object

added modifications to the README file too


core/src/db/seat_assignment/base_seat_assignment.entity.ts
```diff
@@ -0,0 +1,12 @@
...
+import { SeatEntity } from '../seat/seat.entity';
+import { OneToOne } from 'typeorm';
+import { UserEntity } from '../user/user.entity';
+
+export abstract class BaseSeatAssignmentEntity extends BaseEntity {
+	@OneToOne(() => SeatEntity)
+	seat: Promise<SeatEntity>;
+
+	@OneToOne(() => UserEntity)
+	user: Promise<UserEntity>;
```
gankoji, line 11: I'm still not convinced that User is always correct here. We can revisit later, though.

core/src/gql/club/club.object.ts
```diff
@@ -1,13 +1,21 @@
...
+import { FranchiseObject } from '../franchise/franchise.object';
+import { GameObject } from '../game/game.object';
+import { BaseObject } from '../base.object';
 
 @ObjectType('Club')
-export class ClubObject {
-  @Field()
-  id: string;
+export class ClubObject extends BaseObject {
+	// Field is implicit because of ResolveField
```
gankoji, line 9: ResolveField isn't included in this file. Is this comment accurate?

core/src/gql/franchise/franchise.object.ts
```diff
@@ -1,16 +1,12 @@
...
-  @Field()
-  createdAt: Date;
-
-  @Field()
-  updateAt: Date;
-
-  @Field()
-  franchise_name: string;
+	// Field is implicit because of ResolveField
+	club: ClubObject;
```
gankoji, line 11: OneToMany is necessary here.

gankoji: Sorry, that's DB/ORM parlance. We just need club to be `clubs: ClubObject[]`

gankoji (approved):

core/src/gql/club/club.object.ts
matthieu-trudel: that's my bad, i had implemented it in the resolver class, but didn't do a PR for the resolver classes yet
I can change this to @Field for now and overwrite it when I push the PR for the resolver classes

PR #576, opened 2025-06-13
Fix for circular dependencies

matthieu-trudel:
Changes:
datasource.ts: line 44-49 (the rest is whitespace) 
a for loop going through all these entities to initialize them when running the datasource scripts

db.module is now loading from internal.ts instead of the other entities

internal.ts: the file that holds all the entity classes; this will allow us to import all entities from this file to avoid circular dependencies. a more descriptive comment is at the top of the file


core/src/db/db.module.ts
```diff
@@ -2,94 +2,127 @@ import { Module } from '@nestjs/common';
...
-import { UserRepository } from './user/user.repository';
-import { UserAuthAccountEntity } from './user_auth_account/user_auth_account.entity';
-import { UserAuthAccountRepository } from './user_auth_account/user_auth_account.repository';
+	GameEntity,
+	GameModeEntity,
+	RoleEntity,
+	SeatEntity,
+	SkillGroupEntity,
+	FranchiseEntity,
+	//FranchiseSeatAssignmentEntity,
```
gankoji, line 13: Don't leave dead code in.

core/src/db/internal.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * This file is to bypass circular dependencies by initializing the entity classes individually ahead of time.
+ *  ie. Calling a TeamEntity without having initialized ClubEntity beforehand would cause an initialization issue, and vice versa.
+ *
+ * Consequently, we need to import all db entities/repositories from this file
+ */
+export * from './base.entity';
+export * from './user/user.entity';
+//export * from './user/user.seed';
```
gankoji, line 9: Dead code.

gankoji (approved):

PR #772, opened 2026-06-26
fix: revert submission service response to delegate to Zod

gon955:
## Summary

- Seen on two of the mailbox tickets from Jgarret779 and warstog, when submitting they get an unexpected response type object error. 
- Cause: Replacing outputScehma.parse(response) with a large try catch block, which after looking at previous commits was done by Hermes Agents. Would throw an unexpected response type: object error whenever a non string array object was returned e.g CanSubmitReplays.
- According to claude this revert is safe due to the implementation of toJsonbParam() fixing the issue that the hermes agents were tackling to begin with.

## Tests
- Unit test added in 'submission-service-response-handling.spec.ts' that mocks the submission service
- Tests for 3 behaviours: Zod rejects bad shapes, CanSubmitReplays returns the correct object shape and the service sees it as a success, SubmitReplays returns a string array of tasks id's and the service sees it.
- Ran the tests with the old implementation , which fails and returns the unexpected response type:object error
- Ran them with the zod based implementation and they all pass.

### DIFF
--- a/common/src/service-connectors/submission/submission.service.ts
+++ b/common/src/service-connectors/submission/submission.service.ts
@@ -38,31 +38,11 @@ export class SubmissionService {
 
             const response = (await lastValueFrom(rx)) as unknown;
 
-            // Debug: log the raw response type to help diagnose issues
-            this.logger.debug(`Raw response for ${endpoint}: ${JSON.stringify(response).substring(0, 200)}`);
-
-            // Handle case where response might be wrapped or incorrect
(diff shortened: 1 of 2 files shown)


gankoji:
If this is true:
> According to claude this revert is safe due to the implementation of toJsonbParam() fixing the issue that the hermes agents were tackling to begin with.

Then this revert isn't necessary. If it isn't, then the revert isn't safe.

